PR #15, opened 2015-05-08
Adding improvements to assert to support floats and booleans

jeeyoungk:
@achow


achow:
I think EqFloat would be a better name than EqNan, but otherwise LGTM.

jeeyoungk:
Updated!

PR #16, opened 2015-05-08
Adding types to support backend data fetching

jeeyoungk:
* Adding `Backend` interface, which allows you to fetch data.
* Adding time series results and time range structs.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -115,3 +115,107 @@ type TaggedMetric struct {
 
 // GraphiteMetric is a flat, dot-separated identifier to a series of metric.
 type GraphiteMetric string
+
+// Column is a different aspect of data.
+// For example, Blueflood may stores (min / max / average / count) during rollups,
+// and these data are exposed via columns
+type Column string
+
+// Timerange represents a range of time a given time series is defined in:
+// it is 3-tuple of (start, end, resolution) with the following constraints:
+// start <= end
+// start = 0 mod resolution
+// end =   0 mod resolution
+type Timerange struct {
+	Start      int64
+	End        int64
+	Resolution int64
+}
+
+// IsValid determines whether the given timerange meets the constraint.
+func (tr Timerange) IsValid() bool {
+	return (tr.Start%tr.Resolution == 0 &&
+		tr.End%tr.Resolution == 0 &&
+		tr.Resolution > 0 &&
+		tr.Start <= tr.End)
+}
+
+// Slots represent the total # of data points
+func (tr Timerange) Slots() int {
+	return int((tr.Start - tr.End) / tr.Resolution)
+}
+
+// Timeseries is a single time series, identified with the associated tagset.
+type Timeseries struct {
+	Values []float64
+	TagSet TagSet
+}
+
+// SamplingStrategy determines how the given time series should be sampled
+type SamplingStrategy int
+
+const (
+	// SamplingMax chooses the maximum value.
+	SamplingMax SamplingStrategy = iota + 1
+	// SamplingMin chooses the minimum value.
+	SamplingMin
+	// SamplingMean chooses the average value.
+	SamplingMean
+)
+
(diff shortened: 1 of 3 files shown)


jeeyoungk:
CC: @achow

api/types.go
```diff
@@ -115,3 +115,108 @@ type TaggedMetric struct {
 
 // GraphiteMetric is a flat, dot-separated identifier to a series of metric.
 type GraphiteMetric string
+
+// Column is a different aspect of data.
+// For example, Blueflood may stores (min / max / average / count) during rollups,
+// and these data are exposed via columns
+type Column string
```
achow: Sounds like this should be something like a SeriesType? Column seems too generic and backend-aware.

api/types.go
```diff
@@ -115,3 +115,108 @@ type TaggedMetric struct {
...
+
+// SamplingStrategy determines how the given time series should be sampled.
+// Note(This is currently unused).
+type SamplingStrategy int
+
+const (
+	// SamplingMax chooses the maximum value.
+	SamplingMax SamplingStrategy = iota + 1
+	// SamplingMin chooses the minimum value.
+	SamplingMin
```
achow: Don't you need iota here?

api/types.go
jeeyoungk: sure.

api/types.go
jeeyoungk: i think iota preserves itself (magically)

https://splice.com/blog/iota-elegant-constants-golang/

achow: Oh magic. So what's the +1 above for?

jeeyoungk: I think by default it `iota` starts at 0 - i wanted to the values to start at 1 instead.

jeeyoungk: https://play.golang.org/p/VyxirSn4wR

seems like it is the behavior i want.

api/types.go
```diff
@@ -115,3 +115,108 @@ type TaggedMetric struct {
...
+
+// Sample converts the given serieslist to comform with the provided sampling strategy.
+func (s SeriesList) Sample(timerange Timerange, sampling SamplingStrategy) SeriesList {
+	// TODO - deal with the different range.
+	return s
+}
+
+// ScalarSeriesList represents a scalar value returned by the query.
+// such scalar value can be converted to a SeriesList of any timerange.
+type ScalarSeriesList float64
```
achow: We discussed, we can probably get rid of the special typing for "scalar series". We can create the series when we're building the query and just operate on it as if it were any other series, since any sample operation on a scalar series list ends up being the same.

api/types.go
```diff
@@ -115,3 +115,108 @@ type TaggedMetric struct {
...
+		values[i] = value
+	}
+	return SeriesList{
+		List:      []Timeseries{Timeseries{values, NewTagSet()}},
+		Timerange: timerange,
+	}
+}
+
+// MetricMetadata is metadata associated with the given metric.
+type MetricMetadata struct {
```
achow: Would this be more useful as { seriesName : { meta_key : meta_value } }? Or otherwise passing this information closer to what it's supposed to be metadata for? Parallel arrays is a bit weird.

api/types.go
jeeyoungk: done.

jeeyoungk:
updated

api/types.go
```diff
@@ -115,3 +115,94 @@ type TaggedMetric struct {
...
+	return true // validation is now successful.
+}
+
+// Sample converts the given serieslist to comform with the provided sampling strategy.
+func (list SeriesList) Sample(timerange Timerange, sampling SamplingStrategy) SeriesList {
+	// TODO - deal with the different range.
+	return list
+}
+
+// MetricMetadata is metadata associated with the given metric.
```
achow: Should clarify in comments that this is metadata for metrics in the backend. Maybe this belongs in api/backend?

PR #18, opened 2015-05-12
updating the grammar to store the group bys in the function invocations

jeeyoungk:
@achow

### DIFF
--- a/query/language.peg
+++ b/query/language.peg
@@ -101,6 +101,9 @@ expression_3 <-
   <NUMBER> { p.addNumberNode(buffer[begin:end]) }
 
 expression_function  <-
+  # We allow syntax of the form:
+  # func(expr_a, expr_b, expr_c group by column_a, column_b, column_c)
+  # a single optional group-by clause.
   <IDENTIFIER> {
     p.addStringLiteral(unescapeLiteral(buffer[begin:end]))
   }
--- a/query/language.peg
+++ b/query/language.peg
@@ -105,7 +105,7 @@ expression_function  <-
     p.addStringLiteral(unescapeLiteral(buffer[begin:end]))
   }
   PAREN_OPEN
-    expressionList (__ groupByClause)?
+    expressionList { p.addGroupBy() } (__ groupByClause)?
   PAREN_CLOSE {
     p.addFunctionInvocation()
   }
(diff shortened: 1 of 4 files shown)


query/language.peg
```diff
@@ -105,7 +105,7 @@ expression_function  <-
     p.addStringLiteral(unescapeLiteral(buffer[begin:end]))
   }
   PAREN_OPEN
-    expressionList (__ groupByClause)?
+    expressionList (__ groupByClause / { p.addGroupBy() } )
```
achow: Since the addGroupby() is common in both paths, can we make this `expressionList { p.addGroupBy() } (__ groupByClause)?`, and pull the addGroupBy out of groupByClause?

It also looks like the `group by` in a function is going to apply over all arguments?

query/language.peg
```diff
@@ -118,8 +118,15 @@ expression_metric <-
...
   }
 
-groupByClause <-
-  "group" __ "by" __ COLUMN_NAME ( COMMA COLUMN_NAME )*
+groupByClause <- { p.addGroupBy() }
+  "group" __ "by" __ <COLUMN_NAME> {
+    p.appendGroupBy(unescapeLiteral(buffer[begin:end]))
+  }
+  (
+    COMMA COLUMN_NAME {
```
achow: Does this need <> for capturing?

query/language.peg
jeeyoungk: I can! the reason why I've done this is because I wanted to make `groupByClause` to be self-containing. i.e. don't depend anything on the stack for it.

it also should only apply to the last "argument" (not really an argument, but something that gets passed along with everything).

jeeyoungk: ok fixed now.

query/language.peg
jeeyoungk: yes. that's my mistake.

query/language.peg
```diff
@@ -105,7 +105,7 @@ expression_function  <-
     p.addStringLiteral(unescapeLiteral(buffer[begin:end]))
   }
   PAREN_OPEN
-    expressionList (__ groupByClause)?
+    expressionList { p.addGroupBy() } (__ groupByClause)?
```
achow: So are `group by`s meant to apply over all expressions within an expressionList?

`x, y group by z` == `x group by z, y group by z`?

jeeyoungk: added comment.

achow:
LGTM

PR #21, opened 2015-05-19
changed the range clause to the concept of properties

jeeyoungk:
As discussed, I changed the `range` clause to support (somewhat arbitrary) set of key-value pairs. nothing uses this yet, so only the grammar is changed. PTAL!

### DIFF
--- a/query/language.peg
+++ b/query/language.peg
@@ -203,7 +203,7 @@ PROPERTY_KEY <-
   "from" /
   "to" /
   "resolution" /
-  "sample"
+  "sample" __ "by"
 
 PROPERTY_VALUE <- TIMESTAMP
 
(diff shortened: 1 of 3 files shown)


jeeyoungk:
CC: @achow

query/language.peg
```diff
@@ -198,21 +198,29 @@ ID_SEGMENT <-  ID_START ID_CONT*
 # If hyphens are needed, use backticks instead.
 ID_START <-    [a-zA-Z_]
 ID_CONT <-     ID_START / [0-9]
+
+PROPERTY_KEY <-
+  "from" /
+  "to" /
+  "resolution" /
+  "sample"
```
achow: Maybe "sample by" or "sample with"?

achow:
LGTM

PR #23, opened 2015-05-26
Add flag to interpolateTags to check all tags in tagList are used

Nathan-Fenner:
…(Rather than only a proper subset) are used in the pattern; add tests corresponding in rules_test.go to reflect this change

### DIFF
--- a/internal/rules.go
+++ b/internal/rules.go
@@ -16,7 +16,6 @@ package internal
 
 import (
 	"bytes"
-	"errors"
 	"regexp"
 	"strings"
 
@@ -318,9 +317,11 @@ func interpolateTags(pattern string, tagSet api.TagSet, enforceAllTagsUsed bool)
 		}
 	}
 
-	for key := range tagSet {
-		if !usedTags[key] && enforceAllTagsUsed {
-			return "", errors.New("unused key " + key + " in tagset")
+	if enforceAllTagsUsed {
+		for key := range tagSet {
+			if !usedTags[key] {
+				return "", newUnusedTag(key)
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/rules.go
```diff
@@ -315,5 +317,12 @@ func interpolateTags(pattern string, tagSet api.TagSet) (string, error) {
 			buffer.WriteString(token)
 		}
 	}
+
+	for key := range tagSet {
+		if !usedTags[key] && enforceAllTagsUsed {
```
achow: Should pull the test for enforceAllTagsUsed outside of the for.

internal/rules_test.go
```diff
@@ -210,3 +210,48 @@ func TestToGraphiteName_Error(t *testing.T) {
...
+
+func Test_interpolateTags(t *testing.T) {
+
+	for _, testCase := range []struct {
+		pattern string
+		tagSet  api.TagSet
+		enforce bool
+		result  string
+	}{
+		// note that the result <fail> indicates that the test case should fail to parse
```
achow: Should use another flag to indicate whether the test should succeed or fail instead of a special `<fail>` string

internal/rules.go
```diff
@@ -315,5 +317,12 @@ func interpolateTags(pattern string, tagSet api.TagSet) (string, error) {
...
+			return "", errors.New("unused key " + key + " in tagset")
```
jeeyoungk: this is more of a nit, but let's ensure a subtype of `ConversionError` is used. See `newMissingTag` in `errors.go`

jeeyoungk:
LGTM

PR #26, opened 2015-05-28
changing from metric to tagset

jeeyoungk:
As discussed today, i'm changing the `MetricKey` to `TagSet` and moving the human-readable name to `SeriesList`.

@achow @Nathan-Fenner

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -154,7 +154,7 @@ func (tr Timerange) Slots() int {
 // Timeseries is a single time series, identified with the associated tagset.
 type Timeseries struct {
 	Values []float64
-	Metric TaggedMetric
+	TagSet TagSet
 }
 
 // SampleMethod determines how the given time series should be sampled.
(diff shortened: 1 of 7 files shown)


api/types.go
```diff
@@ -154,7 +154,7 @@ func (tr Timerange) Slots() int {
 // Timeseries is a single time series, identified with the associated tagset.
 type Timeseries struct {
 	Values []float64
-	Metric TaggedMetric // TODO (jee) - change this to use TagSet instead of TaggedMetric.
+	TagSet TagSet
```
jeeyoungk: this is the change i want to have - everything else is there to support this change.

api/types.go
```diff
@@ -174,6 +174,7 @@ const (
 type SeriesList struct {
 	Series    []Timeseries
 	Timerange Timerange
+	Name      string // human-readable description of the given time series.
```
jeeyoungk: not being used in many different places - but we should be using it in future.

achow:
A bit surprised that Name's equivalent isn't actually used anywhere. LGTM though.

Nathan-Fenner:
LGTM too

PR #27, opened 2015-05-28
implementation and testing of joins

Nathan-Fenner:
@jeeyoungk @achow

### DIFF
--- a/query/join.go
+++ b/query/join.go
@@ -1,0 +1,86 @@
+// Copyright 2015 Square Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package query
+
+import (
+	"github.com/square/metrics/api"
+)
+
+type JoinRow struct {
+	// The tagSet is used to improve performance, or, possibly in the future, for later queries
+	TagSet api.TagSet
+	// The Row consists of all Timeseries which got collected into this JoinRow
+	Row []api.Timeseries
+}
+
+type JoinResult struct {
+	Rows []JoinRow
+}
+
+// This method takes a partial joinrow, and evaluates the validity of appending `series` to it.
+// If this is possible, return the new series and true; otherwise return false for "ok"
+func extendRow(row JoinRow, series api.Timeseries) (JoinRow, bool) {
+	for key, newValue := range series.Metric.TagSet {
+		oldValue, ok := map[string]string(row.TagSet)[key]
+		if ok && newValue != oldValue {
+			// If this occurs, then the candidate member (series) and the rest of the row are in
+			// conflict about `key`, since they assign it different values. If this occurs, then
+			// it is not possible to assign any key here.
+			return JoinRow{}, false
+		}
+	}
+	// if this point has been reached, then it is possible to extend the row without conflict
+	newTagSet := make(map[string]string)
+	result := JoinRow{newTagSet, append(row.Row, series)}
+	for key, newValue := range series.Metric.TagSet {
+		newTagSet[key] = newValue
+	}
+	for key, oldValue := range row.TagSet {
+		newTagSet[key] = oldValue
+	}
+	return result, true
+}
+
+func Join(lists []api.SeriesList) JoinResult {
+	// place an empty row inside the results list first
+	// this row will be used to build up all others
+	emptyRow := JoinRow{make(map[string]string), []api.Timeseries{}}
+	results := []JoinRow{emptyRow}
+
+	// The `results` list is given an inductive definition:
+	// at the end of the `i`th iteration of the outer loop,
+	// `results` corresponds to the Join of the first `i` seriesLists given as input
+
+	for _, list := range lists {
+		next := []JoinRow{}
+		// `next` is gradually accumulated into the final Join of the first `i` seriesLists
+		// so that at the end of the loop, we can assign:
+		//     results = next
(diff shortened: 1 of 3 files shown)


query/join.go
```diff
@@ -0,0 +1,86 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package query
+
+import (
+	"github.com/square/metrics/api"
+)
+
+type JoinRow struct {
```
jeeyoungk: i think `JoinRow`, `Join`, and `JoinResult` can all start with lowercase, because I don't think it needs to be used outside the package.

(go uses the capitalization of the first character to determine whether a given symbol is exported or not).

query/join.go
```diff
@@ -0,0 +1,86 @@
...
+		oldValue, ok := map[string]string(row.TagSet)[key]
+		if ok && newValue != oldValue {
+			// If this occurs, then the candidate member (series) and the rest of the row are in
+			// conflict about `key`, since they assign it different values. If this occurs, then
+			// it is not possible to assign any key here.
+			return JoinRow{}, false
+		}
+	}
+	// if this point has been reached, then it is possible to extend the row without conflict
+	newTagSet := make(map[string]string)
```
jeeyoungk: `api.NewTagSet()`

query/join_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Fatalf("Join of a row with itself changes its length")
+	}
+}
+
+func Test_Join_CartesianProduct(t *testing.T) {
+	// attempt to join two different sets with nothing in common
+	// and check that they multiply
+	result := Join([]api.SeriesList{basicList, envList})
+	if len(result.Rows) != len(basicList.Series)*len(envList.Series) {
+		t.Fatalf("Join with serieslists having disjoint key sets results in unexpected number of entries (should be %v but is %v)", len(basicList.Series)*len(envList.Series), len(result.Rows))
```
jeeyoungk: let's use `%d` for formatting values known to be integers.

query/join_test.go
```diff
@@ -0,0 +1,127 @@
...
+// to modifications to the series lists used as example testcases
+func max(x, y int) int {
+	if x < y {
+		return y
+	} else {
+		return x
+	}
+}
+
+func Test_Join_KeySubsets(t *testing.T) {
```
jeeyoungk: i think i might combine both tests (this `Join_KeySubsets` and `Join_Single`), and do table-based testing. i.e. store the test suites (input + output pair) in an array, and simply iterate through them. you can look at files in `query_test.go` or other files for more info.

query/join_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// and check that they multiply
+	result := Join([]api.SeriesList{basicList, envList})
+	if len(result.Rows) != len(basicList.Series)*len(envList.Series) {
+		t.Fatalf("Join with serieslists having disjoint key sets results in unexpected number of entries (should be %v but is %v)", len(basicList.Series)*len(envList.Series), len(result.Rows))
+	}
+}
+
+// A simple max function used in the Test_Join_KeySubsets which makes it less brittle
+// to modifications to the series lists used as example testcases
+func max(x, y int) int {
```
jeeyoungk: think no comment is necessary (pretty obvious) - and let's move it to the bottom of the file.

query/join_test.go
```diff
@@ -0,0 +1,127 @@
...
+	if len(result.Rows) != 0 {
+		t.Fatalf("Join of single (nonempty row, empty row) produces non-empty result")
+	}
+	result = Join([]api.SeriesList{basicList, basicList, basicList, emptyList, basicList})
+	if len(result.Rows) != 0 {
+		t.Fatalf("Join of many nonempty rows and one empty row produces non-empty result")
+	}
+}
+
+func Test_Join_Self(t *testing.T) {
```
jeeyoungk: i would make this table-based testing also (see the other comment).

query/join.go
```diff
@@ -0,0 +1,86 @@
...
+	for key, oldValue := range row.TagSet {
+		newTagSet[key] = oldValue
+	}
+	return result, true
+}
+
+func Join(lists []api.SeriesList) JoinResult {
+	// place an empty row inside the results list first
+	// this row will be used to build up all others
+	emptyRow := JoinRow{make(map[string]string), []api.Timeseries{}}
```
jeeyoungk: let's call `api.newTagSet()` here also.

query/join.go
```diff
@@ -0,0 +1,86 @@
...
+	results := []JoinRow{emptyRow}
+
+	// The `results` list is given an inductive definition:
+	// at the end of the `i`th iteration of the outer loop,
+	// `results` corresponds to the Join of the first `i` seriesLists given as input
+
+	for _, list := range lists {
+		next := []JoinRow{}
```
jeeyoungk: is this variable needed? feels like `results` and `next` has the same role. i think not having this variable and just using `results` everywhere should be sufficient?

query/join.go
```diff
@@ -0,0 +1,86 @@
...
+	for key, newValue := range series.Metric.TagSet {
+		newTagSet[key] = newValue
+	}
+	for key, oldValue := range row.TagSet {
+		newTagSet[key] = oldValue
+	}
+	return result, true
+}
+
+func Join(lists []api.SeriesList) JoinResult {
```
jeeyoungk: maybe a little comment on what we're trying to do here - `generates a cartesian product of the provided arrays, and emits rows where the tags are matching`.

query/join.go
Nathan-Fenner: `results` is ranged over and is unchanged within each iteration of the loop (until being assigned to `next`).

To do this with only one variable would require stopping the `range` once the next iteration's values are encountered, then dropping all values prior to that point from the slice.

query/join.go
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+
+package query
+
+import (
+	"github.com/square/metrics/api"
+)
+
+type joinRow struct {
+	// The tagSet is used to improve performance, or, possibly in the future, for later queries
```
jeeyoungk: comment should be placed along with the field
`name type //comment`

jeeyoungk:
LGTM with nit.

PR #28, opened 2015-05-28
Nathanf join with binary expressions

Nathan-Fenner:
@jeeyoungk  @achow 
Uses the join operation to compute binary functions (like the operators +, -, *, /).

### DIFF
--- a/query/expression_test.go
+++ b/query/expression_test.go
@@ -244,6 +244,56 @@ func Test_evaluateBinaryOperation(t *testing.T) {
 			true,
 			[][]float64{{8, 10, 12}, {14, 16, 18}, {12, 15, 18}, {21, 24, 27}},
 		},
+		{
+			emptyContext,
+			"add",
+			[]Expression{
+				&LiteralSeriesExpression{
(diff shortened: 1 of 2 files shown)


query/expression_test.go
```diff
@@ -127,7 +142,157 @@ func Test_evaluateBinaryOperation(t *testing.T) {
...
+							api.TagSet{
+								"env": "staging",
+							},
+						},
+					},
+				},
+			},
+			func(left, right float64) float64 { return left * right },
+			true,
+			[][]float64{{8, 10, 12}, {14, 16, 18}, {12, 15, 18}, {21, 24, 27}},
```
jeeyoungk: so right now, we just drop `env=production:host=#1`

Nathan-Fenner: The tags do persist (see line 137 in query/expression.go), this test just does not check for them. It could be a good idea to add that information to the test.

jeeyoungk: i see. i think it's fine.

jeeyoungk:
LGTM

PR #30, opened 2015-05-28
DNM Set up travis ci

achow:


jeeyoungk:
LGTM! OMG so excited.

we should add linters to this when we figure out how to (i think we should make a `Makefile` for it because we should avoid auto-generated files.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: go
+go: 1.4
+
+services:
+  - cassandra
```
jeeyoungk: so simple.

PR #35, opened 2015-06-01
Aggregation Function Implementation

Nathan-Fenner:
This pull request defines and implements aggregator functions which can be used to take a list of timeseries, group them by their tags, and then combine each group into a single timeseries using an aggregator like their sum, minimum, maximum, or mean.

`func aggregateBy(list api.SeriesList, aggregator func([]float64) float64, tags []string) (api.SeriesList, error)` is the API for access by the rest of the query system, along with the aggregator struct types `sumAggregator`, `meanAggregator`, `maxAggregator`, and `minAggregator`.

@jeeyoungk @achow

### DIFF
--- a/query/aggregate.go
+++ b/query/aggregate.go
@@ -28,13 +28,9 @@ type group struct {
 	TagSet api.TagSet
 }
 
-type groupResult struct {
-	Results []group
-}
-
 // If the given group will accept this given series (since it belongs to this group)
-// then bucketValid will return true.
-func bucketValid(left api.TagSet, right api.TagSet, tags []string) bool {
+// then groupAccept will return true.
+func groupAccepts(left api.TagSet, right api.TagSet, tags []string) bool {
 	for _, tag := range tags {
 		if left[tag] != right[tag] {
 			return false
@@ -44,7 +40,7 @@ func bucketValid(left api.TagSet, right api.TagSet, tags []string) bool {
 }
 
 // Adds the series to the corresponding bucket, possibly modifying the input `rows` and returning a new list.
-func addToBucket(rows []group, series api.Timeseries, tags []string) []group {
+func addToGroup(rows []group, series api.Timeseries, tags []string) []group {
 	// First we delete all tags with names other than those found in 'tags'
 	newTags := api.NewTagSet()
 	for _, tag := range tags {
@@ -55,7 +51,7 @@ func addToBucket(rows []group, series api.Timeseries, tags []string) []group {
 
 	// Next, find the best bucket for this series:
 	for i, row := range rows {
-		if bucketValid(row.TagSet, series.TagSet, tags) {
+		if groupAccepts(row.TagSet, series.TagSet, tags) {
 			rows[i].List = append(rows[i].List, series)
 			return rows
 		}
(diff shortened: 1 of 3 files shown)


query/aggregate.go
```diff
@@ -0,0 +1,210 @@
...
+	return max
+}
+
+// applyAggregation takes an aggregation function ( [float64] => float64 ) and applies it to a given list of Timeseries
+// the list must be non-empty, or an error is returned
+func applyAggregation(group group, aggregator aggregator) (api.Timeseries, error) {
+	list := group.List
+	tagSet := group.TagSet
+
+	if len(list) == 0 {
```
achow: Why enforce non-empty list of Timeseries?

query/errors.go
```diff
@@ -43,4 +43,13 @@ func (errors SyntaxErrors) Error() string {
 	return strings.Join(errorStrings, "\n")
 }
 
+// EmptyAggregateError is an Error for attempts to aggregate empty SeriesLists.
+type EmptyAggregateError struct {
```
achow: Since there's nothing special for this error, you can just create it using errors.New()

query/aggregate.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return series, nil
+}
+
+// This function is the culmination of all others.
+// `aggregateBy` takes a series list, an aggregator, and a set of tags.
+// It produces a SeriesList which is the result of grouping by the tags and then aggregating each group
+// into a single Series.
+func aggregateBy(list api.SeriesList, aggregator aggregator, tags []string) (api.SeriesList, error) {
```
achow: We should have callers use an enum instead of an actual `aggregator` when calling into this. `aggregators` end up being just an `aggregate()`, so we can key the enum into a hash of `aggregate()` functions.

query/aggregate.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"github.com/square/metrics/api"
+)
+
+type group struct {
+	List   []api.Timeseries
+	TagSet api.TagSet
+}
+
+type groupResult struct {
```
achow: Looks like this is unused.

query/aggregate.go
```diff
@@ -0,0 +1,210 @@
...
+	Results []group
+}
+
+// If the given group will accept this given series (since it belongs to this group)
+// then bucketValid will return true.
+func bucketValid(left api.TagSet, right api.TagSet, tags []string) bool {
```
achow: Should standardize on `group` instead of `bucket`. This is more of a compatibility check between tags, so I think we'd benefit from naming this something along those lines.

query/aggregate.go
Nathan-Fenner: I'm not sure how exactly to address it.

On one hand, if the group is empty, then there is no way to guess information like the number of values, or the time range, or the sampling rate, etc.

On the other hand, the function is never called with an empty group, since every group created by the `groupBy` gets at least one list put inside it.

So the error checking (and the returned error) could be removed entirely... but in the future, if a bug was introduced (or the specification changed) in `groupBy` or a related function, then a panic will occur here where `list[0]` is chosen to find the correct length for the output `Values`.

query/aggregate.go
Nathan-Fenner: It is, will delete.

query/aggregate.go
achow: My feel is that aggregates on an empty list of Timeseries should just return an empty list. @jeeyoungk ?

query/aggregate.go
```diff
@@ -0,0 +1,224 @@
...
+type aggregate int
+
+const (
+	sumAggregate aggregate = iota + 1
+	meanAggregate
+	minAggregate
+	maxAggregate
+)
+
+var aggregatorMap = map[aggregate]aggregator{
```
achow: Meant more of `map[aggregate]func([]float64)float64`. There's no need to go through a struct since these are all just different aggregate functions.

Nathan-Fenner: Oh, I see, that makes sense.

query/aggregate.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return series
+}
+
+// This function is the culmination of all others.
+// `aggregateBy` takes a series list, an aggregator, and a set of tags.
+// It produces a SeriesList which is the result of grouping by the tags and then aggregating each group
+// into a single Series.
+func aggregateBy(list api.SeriesList, aggregate aggregate, tags []string) api.SeriesList {
```
jeeyoungk: i think it's better if we pass in `func([]float64)float64` directly from the outside. they can do the enum-to-function mapping.

(one better thing about that approach is that we can make this file more generic, and additional functions can be added in tests, etc without modifying this file).

query/aggregate.go
jeeyoungk: I looked at the code. I actually think panicking isn't too bad because we the grouping code shouldn't ever generate a group with an empty list, right?

jeeyoungk: regardless it seems (at this point - I can't think of a good case) that it's a bit silly to discuss about a code path that we don't really know how it's going to be executed. panicking might be better because at least we'll be aware of this path being hit immediately. I would suggest even explicitly calling `panic` here.

query/aggregate_test.go
```diff
@@ -0,0 +1,451 @@
...
+		aggregatorMap[maxAggregate],
+		[]float64{4, 2, 4, 4},
+	},
+	{
+		aggregatorMap[minAggregate],
+		[]float64{-1, -1, 0, 2},
+	},
+}
+
+const epsilon = 1e-10 // epsilon is a constant for the maximum allowable error between correct test case answers and actual results
```
jeeyoungk: let's move this to the top.

query/aggregate.go
achow: True. I wanted the enum so we could have a common place for callers that are outside of the query engine that would want to do aggregations on their own. We could have `AggregateBy()` take the enum and just call into an `AggregateByFunc()` to allow both behaviors. Either way works for me though, just something to think about.

query/aggregate_test.go
```diff
@@ -0,0 +1,451 @@
...
+package query
+
+import (
+	"github.com/square/metrics/api"
+
+	"math"
+	"testing"
+)
+
+var (
```
jeeyoungk: AFAIK, all the test suites in the file scope in this file are independent of each other, right? In that case, let's move them into each of the functions instead - it's a bit hard to read because i need to jump between the test case and the test function they're being used in.

jeeyoungk:
I think it's good for now. Try squashing them to a single commit before merging to give the perception to the entire world that we write perfect codes in one shot. :P

jeeyoungk:
LGTM with the last nit.

achow:
I think we should export the `aggregate` enums and `aggregateBy()`, but otherwise LGTM.

jeeyoungk:
@achow I wouldn't worry too much about access control within the `query` package because it's the package without any dependency (and everything is package private currently).

Though, if we make this "generic", then you are right - we should export these two functions and enums.

PR #36, opened 2015-06-01
Implement Evaluation Context Assignment and Parsing

Nathan-Fenner:
Currently:
* implemented (initial) evaluation context assignment

Main issues/unexpected changes outside of implementation:
* The PEG grammar had to be modified in several places where it did not act as intended.
* The `SelectCommand` got a new member `context` which holds evaluation context information (but no backend)
* The `SelectCommand.Execute` method had to be modified accordingly
* The parser is very, very brittle when dealing with string literals; escaping characters and the quotes themselves are currently part of the parsed string. "Literal String" parsing should (ideally) substitute the source strings with their unescaped friends. Strings are currently like this: `"'this is a string \\' that has an escaped single quote'"`, which is hard to process.
* Parsing dates is fairly rigid; the current approach is to try one of 6 formats (based on presence/absence of minute/second/hour/timezone) or allow a literal integer to represent a unix timestamp. This may need to be changed.
*A modification to the testing (calling `p.Parse()` instead of `Parse()`) which has been performed in this branch, because the `.Parse()` method does not accurately check for errors.

@jeeyoungk  @achow

### DIFF
--- a/query/node.go
+++ b/query/node.go
@@ -88,9 +88,10 @@ type functionExpression struct {
 // ---------------
 // These nodes are only present during the parsing step and are not present
 // in the resulting command.
-// There are two types of temporary nodes:
+// There are three types of temporary nodes:
 // * literals (constants in the syntax tree).
 // * lists
+// * evaluation context nodes
 
 type stringLiteral struct {
 	literal string
@@ -117,6 +118,23 @@ type expressionList struct {
 
 type groupByList struct {
 	list []string
+}
+
+// evaluationContextKey represents a key (from, to, sampleby) for the evaluation context.
+type evaluationContextKey struct {
+	key string
+}
+
+// evaluationContextValue represents a value (date, samplingmode, etc.) for the evaluation context.
+type evaluationContextValue struct {
+	value string
+}
+
+// evaluationContextMap represents a collection of key-value pairs that form the evaluation context.
+type evaluationContextNode struct {
+	Timerange    api.Timerange    // Timerange to fetch data from
+	SampleMethod api.SampleMethod // to use when up/downsampling to match requested resolution
+	assigned     map[string]bool  // a map for knowing which elements of the context have been assigned
 }
 
 // Helper functions for printing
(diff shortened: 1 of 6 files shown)


query/node.go
```diff
@@ -88,9 +88,10 @@ type functionExpression struct {
 // ---------------
 // These nodes are only present during the parsing step and are not present
 // in the resulting command.
-// There are two types of temporary nodes:
+// There are three types of temporary nodes:
 // * literals (constants in the syntax tree).
 // * lists
+// * evaluation context nodes
```
Nathan-Fenner: Although they initially were intended to be, they aren't actually temporary anymore.

query/parser.go
```diff
@@ -213,6 +251,131 @@ func (p *Parser) addOperatorFunction() {
...
+		case "'mean'":
+			contextNode.SampleMethod = api.SampleMean
+		default:
+			p.flagSyntaxError(SyntaxError{
+				token:   value,
+				message: fmt.Sprintf("Expected sampling method 'max', 'min', or 'mean'", value),
+			})
+		}
+	case "from":
+
```
jeeyoungk: whitespace. and i think we can combine `from` and `to` case into one, since they do the same thing?

query/query_test.go
```diff
@@ -89,27 +90,36 @@ var parseOnly = []string{
...
 	for _, row := range inputs {
-		if err := checkSyntaxError(t, row); err != nil {
+		_, err := Parse(row)
+		if err != nil {
 			t.Errorf("[%s] failed to parse: %s", row, err.Error())
 		}
 	}
-	for _, row := range parseOnly {
-		if err := checkSyntaxError(t, row); err != nil {
+	/*for _, row := range parseOnly {
```
jeeyoungk: why did you comment it out?

jeeyoungk:
You may need to run .`/build.sh` again after updating peg.

jeeyoungk:
regarding the string literal parsing - i have to agree. I was kinda lazy when I was implementing it the first time (since you need to have different character classes). I will create a task for making string parsing better.

query/query_test.go
Nathan-Fenner: Originally it broke and made things hard to debug - now it's fine so I put it back

query/node.go
jeeyoungk: I don't think it's too important, since it's only the logical structure of the code (also, key,values are temporary).

jeeyoungk:
LGTM

PR #38, opened 2015-06-02
Travis CI: Golint, Gofmt, and Build.sh checks

Nathan-Fenner:
Now green! Adds checks to the CI build for the grammar file having been built correctly, as well as gofmt having been run.

@jeeyoungk @achow


query/build.sh
```diff
@@ -1,2 +1,3 @@
 #!/usr/bin/env bash
-peg -inline -switch language.peg
+
+peg -inline -switch query/language.peg
```
achow: Let's point use absolute path to language.peg, so it works from any dir.

build_verify.sh
```diff
@@ -0,0 +1,77 @@
...
+	echo "$GOFMT_RESULTS"
+	fails="fails"
+fi
+
+
+# Find all files ending with .go that do not end with .peg.go (generated)
+GOLINT_FILES=`find . -not -name *.peg.go -name *.go`
+# Iterate over these (split by newline)
+
+# Check each of them with golint
```
achow: I don't think we should fail the build on `golint` failures, but we should run them.

jeeyoungk: what's the canonical way of having multi-component build with a possible "warning" state with Travis? otherwise we can just choose to not fail.

sadly we can't return `0.5` as an Unix exit code...

achow: Not sure there is one. I kind of agree with the sentiment that things should either pass/fail and have nothing in between though. If it's something we really want to warn about, we should fail. Especially in the case of golint, we'd perpetually be in a warning state if we never fix it.

Maybe we can pull out the `golint` logic into its own script so we can run it in dev.

.travis.yml
```diff
@@ -8,4 +8,11 @@ before_install:
...
 
 before_script:
+  # golint
+  - go get -u github.com/golang/lint/golint
+  # peg
+  - go get -u github.com/pointlander/peg
   - cqlsh -f schema/schema_test.cql
+
+script:
+  - ./build_verify.sh
```
achow: Should explicitly add `go test -v ./...` here, since otherwise travis assumes that script has everything we want to run in the build.

query/build.sh
Nathan-Fenner: What path would that be, so that it works on all machines?

.travis.yml
Nathan-Fenner: added below

build_verify.sh
```diff
@@ -0,0 +1,48 @@
+#/usr/bin/env bash 
+
+
+fails=""
+
+# Invoke gofmt on the current directory.
+# -l flag causes it to only list files
+GOFMT_RESULTS=`gofmt -l .`
+if [ "$GOFMT_RESULTS" ]
```
jeeyoungk: let's put `] then` into a single line. (i think that's more idiomatic shell scripting)

query/build.sh
jeeyoungk: http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in

`DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )` works. A handy hack.

build_verify.sh
```diff
@@ -0,0 +1,48 @@
...
+	echo "To get the latest version of peg, run:"
+	echo "> go get -u github.com/pointlander/peg"
+	fails="fails"
+fi
+
+if [ $fails ]
+then
+	exit -1
+fi
+
```
jeeyoungk: newlines.

jeeyoungk:
LGTM

PR #41, opened 2015-06-02
Transformation function implementation

Nathan-Fenner:
The interface is:

`func ApplyTransformation(list api.SeriesList, transform transform, parameters []value) (api.SeriesList, error)`

Where `parameters` are `value`s representing things like flags, modes, or durations (and correspond to all arguments after the first in the query)

The type `transform` is a synonym for the more verbose:

`func([]float64, []value, api.Timerange) ([]float64, error)`

The following transformations have been implemented: `transformDerivative`, `transformIntegral`, `transformRate`, `transformCumulative`, and `transformMovingAverage`.

Also implemented is `transformMapMaker` which maps a simple scalar functio `func(float64) float64` over the timeseries. The transform corresponding to taking the absolute value, for example, would be `transformMapMaker(math.Abs)`.

@jeeyoungk @achow

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -36,15 +36,12 @@ func NewTagSet() TagSet {
 }
 
 func (left TagSet) Equals(right TagSet) bool {
+	if len(left) != len(right) {
+		return false
+	}
 	for k := range left {
 		_, ok := right[k]
 		if !ok || left[k] != right[k] {
-			return false
-		}
-	}
-	for k := range right {
-		_, ok := left[k]
-		if !ok {
 			return false
 		}
 	}
--- a/api/types.go
+++ b/api/types.go
@@ -35,6 +35,22 @@ func NewTagSet() TagSet {
 	return make(map[string]string)
 }
 
+func (left TagSet) Equals(right TagSet) bool {
+	for k := range left {
+		_, ok := right[k]
+		if !ok || left[k] != right[k] {
+			return false
+		}
+	}
+	for k := range right {
+		_, ok := left[k]
+		if !ok {
+			return false
+		}
+	}
+	return true
+}
+
 // Merge two tagsets, and return a new tagset.
 // If keys conflict, the first tag set is preferred.
 func (tagSet TagSet) Merge(other TagSet) TagSet {
--- a/query/transformation.go
+++ b/query/transformation.go
@@ -68,7 +68,8 @@ func ApplyTransform(list api.SeriesList, transform transform, parameters []value
 func checkParameters(name string, expected int, parameter transformParameter) error {
 	args := parameter.parameters
 	if len(args) != expected {
-		return errors.New(fmt.Sprintf("Expected function %s to be given %d parameters but was given %d: main series and %+v", name, expected+1, len(args)+1, args))
+		printArgs := append([]value{stringValue("(SeriesList)")}, args...)
+		return errors.New(fmt.Sprintf("expected %s to be given %d parameters but was given %d: %+v", name, expected+1, len(args)+1, printArgs))
 	}
 	return nil
 }
@@ -163,10 +164,11 @@ func transformMovingAverage(values []float64, parameter transformParameter) ([]f
 }
 
 // transformMapMaker can be used to use a function as a transform, such as 'math.Abs' (or similar):
-//  `transformMapMaker(math.Abs)` is a transform function which can be used, e.g. with applyTransform
-func transformMapMaker(fun func(float64) float64) func([]float64, transformParameter) ([]float64, error) {
+//  `transformMapMaker(math.Abs)` is a transform function which can be used, e.g. with ApplyTransform
+// The name is used for error-checking purposes.
+func transformMapMaker(name string, fun func(float64) float64) func([]float64, transformParameter) ([]float64, error) {
 	return func(values []float64, parameter transformParameter) ([]float64, error) {
-		if err := checkParameters("transform.map(???)", 0, parameter); err != nil {
+		if err := checkParameters(fmt.Sprintf("transform.%s", name), 0, parameter); err != nil {
 			return nil, err
 		}
 		result := make([]float64, len(values))
--- a/query/transformation.go
+++ b/query/transformation.go
@@ -17,8 +17,6 @@
 package query
 
 import (
-	"math"
-
 	"github.com/square/metrics/api"
 )
 
@@ -136,21 +134,4 @@ func transformMapMaker(fun func(float64) float64) func([]float64, transformation
 		}
 		return result
 	}
-}
-
-func transformTimeOffset(values []float64, parameter transformationParameter) []float64 {
-	result := make([]float64, len(values))
-	// Shifting the time series by the given number of samples (by dividing the parameter by scale)
-	shift := int(parameter.parameter/parameter.scale + math.Copysign(0.5, parameter.parameter)) // The (+ 0.5) causes it to round.
-	// Positive shift means forward in time (so result[shift] = values[0])
-	// Negative shift means backwards in time.
-	// Values outside this range are assigned 0
-	for i := range values {
-		if i-shift < 0 || i-shift >= len(values) {
-			result[i] = 0
-			continue
-		}
-		result[i] = values[i-shift]
-	}
-	return result
 }
--- a/query/transformation.go
+++ b/query/transformation.go
@@ -23,18 +23,19 @@ import (
 	"github.com/square/metrics/api"
 )
 
-// A transform takes the list of values, and a "scale" factor:
-// It is the unitless quotient (sample duration / 1s)
-// So if the resolution is "once every 5 minutes" them we get a scale factor of 5*60 = 300
-type transform func([]float64, transformParameter) ([]float64, error)
+// A transform takes the list of values, other parameters, and the timerange of the query.
+type transform func([]float64, []value, api.Timerange) ([]float64, error)
 
-type transformParameter struct {
-	scale      float64
-	parameters []value
+// timerangeScale takes an api.Timerange and computes the "scale" or duration of one sample (in seconds).
+// It is useful for transformations that normalize on time (like derivative or integral).
+func timerangeScale(timerange api.Timerange) float64 {
+	return float64(timerange.Resolution)
 }
 
-func transformTimeseries(series api.Timeseries, transform transform, parameter transformParameter) (api.Timeseries, error) {
-	values, err := transform(series.Values, parameter)
+// transformTimeseries transforms an individual series (rather than an entire serieslist) taking the same parameters as a transform,
+// but with the serieslist standing in for the simplified []float64 argument.
+func transformTimeseries(series api.Timeseries, transform transform, parameters []value, timerange api.Timerange) (api.Timeseries, error) {
+	values, err := transform(series.Values, parameters, timerange)
 	if err != nil {
 		return api.Timeseries{}, err
 	}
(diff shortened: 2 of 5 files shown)


query/transformation.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// A transformation takes the list of values, and a "scale" factor:
+// It is the unitless quotient (sample duration / 1s)
+// So if the resolution is "once every 5 minutes" them we get a scale factor of 5*60 = 300
+type transformation func([]float64, transformationParameter) []float64
+
+type transformationParameter struct {
+	scale     float64
+	parameter float64
```
Nathan-Fenner: Might make more sense for the future for this to be either `[]float64` or something that can hold non-float parameters too

jeeyoungk: hmm.

I'm thinking about passing:

* some subset of `EvaluationContext` (everything in it is too much, because we don't want transforms to be doing a lot of crazy things, such as fetching backend. I think `TimeRange` is good to be passed around.

* if the other PR - https://github.com/square/metrics/pull/43 - is merged, maybe we can pass in the list of values?

i like the fact that the first parameter is `[]float64` because it restricts what we can do with it.

Nathan-Fenner: Yes, changing to `[]value` for parameters makes sense with #43 merged.

I'm not sure what a transformation would do with the `Timerange`. Currently they use the `scale` which, as far as I can tell, is all they should care about (and `scale` is the seconds-duration of a single sample, so perhaps poorly named).

query/transformation.go
```diff
@@ -0,0 +1,156 @@
...
+			// So exclude 6 = 10 - 4 = i - limit
+		}
+		result[i] = sum / float64(count)
+	}
+	return result
+}
+
+// transformMapMaker can be used to use a function as a transformation, such as 'math.Abs' (or similar):
+//  `transformMapMaker(math.Abs)` is a transformation function which can be used, e.g. with applyTransformation
+func transformMapMaker(fun func(float64) float64) func([]float64, transformationParameter) []float64 {
```
jeeyoungk: this is nifty.

query/transformation.go
```diff
@@ -0,0 +1,156 @@
...
+	return func(values []float64, parameter transformationParameter) []float64 {
+		result := make([]float64, len(values))
+		for i := range values {
+			result[i] = fun(values[i])
+		}
+		return result
+	}
+}
+
+func transformTimeOffset(values []float64, parameter transformationParameter) []float64 {
```
jeeyoungk: Let's leave out the time shifting for now - I have some other thoughts on handling it.

The biggest issue is that we get bunch of 0s because we don't have enough data. I think the correct way to handle it is by modifying `EvaluationContext` to have a different `TimeRange`. Thus, if you make a (psudocode) call like

```shift(cpu_usage, "2 weeks ago)```

What it would do is - shift doesn't do anything, but changes the evaluation context to fetch 2 weeks earlier data instead. we may need to readjust the returned timerange to ensure that the timeranges are also shifted.

File: query/transformation.go [hunk 4]
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/square/metrics/api"
+)
+
+// A transformation takes the list of values, and a "scale" factor:
+// It is the unitless quotient (sample duration / 1s)
+// So if the resolution is "once every 5 minutes" them we get a scale factor of 5*60 = 300
+type transformation func([]float64, transformationParameter) ([]float64, error)
+
+type transformationParameter struct {
```
achow: Why create a struct for params instead of using them directly?

query/transformation.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2015 Square Inc.
```
achow: I'd say standardize on `transform` instead of `transformation` for naming.

achow:
We should really export the functions that callers are supposed to use. Even if it's only used package-internal.

query/transformation.go
Nathan-Fenner: My concern was that other parameters might be implemented in the future (like providing the `Timerange` explicitly) and if they were given individually then every already-defined transformation which didn't use them would have to change.

If we can set the parameters in stone then it would be a better idea to pass them individually.

query/transformation.go
Nathan-Fenner: Yeah, that's probably better. I've updated it.

query/transformation.go
achow: Either way, you still have to unpack the params you want from the struct to use them. I think we should just provide them directly to the transforms. Seems like this just causes some unnecessary indirection.

query/transformation_test.go
```diff
@@ -0,0 +1,282 @@
...
+			Start:      758300,
+			End:        758300 + 30*5,
+			Resolution: 30,
+		},
+		Name: "test",
+	}
+	testCases := []struct {
+		transform transform
+		parameter []value
+		expected  string
```
achow: This shouldn't compare expected error messages. It's really fragile (tweaking the message slightly means you have to update all these tests). Just testing existence of the error should be sufficient.

query/transformation.go
Nathan-Fenner: The parameters are now unpacked (and the struct type has been removed).

query/transformation_test.go
```diff
@@ -0,0 +1,282 @@
...
+package query
+
+import (
+	"math"
+	"testing"
+
+	"github.com/square/metrics/api"
+)
+
+func tagSetEquals(a, b api.TagSet) bool {
```
achow: I think we use this enough to make this `func (left TagSet) Equals(right TagSet) bool` in api/types.go.

Nathan-Fenner: Okay, implemented in this branch. I found two places where it had been defined; here and in `aggregate_test.go`. I'm not sure where else it might be.

api/types.go
```diff
@@ -35,6 +35,22 @@ func NewTagSet() TagSet {
 	return make(map[string]string)
 }
 
+func (left TagSet) Equals(right TagSet) bool {
```
achow: Needs tests.

query/transformation_test.go
jeeyoungk: +1. I think unless error is structured to a separate type, there's very little value on checking the error value strings.

api/types.go
```diff
@@ -35,6 +35,22 @@ func NewTagSet() TagSet {
...
+	for k := range left {
+		_, ok := right[k]
+		if !ok || left[k] != right[k] {
+			return false
+		}
+	}
+	for k := range right {
```
achow: This part can probably just compare `len(left) == len(right)` and we'd have the same behavior.

jeeyoungk:
LGTM on my side - just looked it over. This is pretty good. Thanks!

api/types.go
Nathan-Fenner: Oh, yeah! Didn't occur to me

query/transformation.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/square/metrics/api"
+)
+
+// A transform takes the list of values, other parameters, and the timerange of the query.
+type transform func([]float64, []value, api.Timerange) ([]float64, error)
+
+// timerangeScale takes an api.Timerange and computes the "scale" or duration of one sample (in seconds).
+// It is useful for transformations that normalize on time (like derivative or integral).
+func timerangeScale(timerange api.Timerange) float64 {
```
achow: If the actual `transform`s only use the timerange resolution, you can make that the parameter instead of shipping around the timerange and having to grab it every time. Then `ApplyTransform` can just pass the resolution along and you have a single place to do whatever fancy stuff you need.

query/transformation.go
```diff
@@ -0,0 +1,195 @@
...
+			return api.SeriesList{}, err
+		}
+	}
+	return result, nil
+}
+
+// checkParameters is used to make sure that each transform is given the right number of parameters.
+func checkParameters(name string, expected int, parameters []value) error {
+	if len(parameters) != expected {
+		printArgs := append([]value{stringValue("(SeriesList)")}, parameters...)
```
achow: What's this append for?

api/types.go
Nathan-Fenner: Tests have been added

query/transformation.go
Nathan-Fenner: Transformation functions are given `[]float64` corresponding to a single timeseries (from a serieslist), in addition to a `parameters []value` which corresponds to all the *other* parameters given in the query. For example, when writing:

`transform.moving_average( cpu, 60 )`

The `parameters` will be the slice `[]value{ 60 }` (ignoring type wrappers).

However, I wanted to try to hide this from the user (since they don't need to care how it's implemented) and they'd like to see the parameters including the time series. The `append` puts a string that says `"(SeriesList)"` to stand for the currently-processing series-list which isn't actually available to the function. Otherwise, when, say, they call `transform.derivative` accidentally with a second argument, it would only show them the second argument.

There might be a better way of doing it, but I'm not familiar with the Go way.

query/transformation_test.go
Nathan-Fenner: The error-message checking has been removed

query/transformation.go
```diff
@@ -0,0 +1,195 @@
...
+		result[i] = (values[i] - values[i-1]) / timerangeScale(timerange)
+		if result[i] < 0 {
+			result[i] = 0
+		}
+	}
+	return result, nil
+}
+
+// transformCumulative computes the cumulative sum of the given values.
+func transformCumulative(values []float64, parameters []value, timerange api.Timerange) ([]float64, error) {
```
achow: This doesn't seem necessary. You can just use `transformIntegral` with a resolution of 1. Is there some use case for this specific behavior regardless of resolution?

query/transformation.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		// Otherwise, it's the scaled difference
+		result[i] = (values[i] - values[i-1]) / timerangeScale(timerange)
+	}
+	return result, nil
+}
+
+// transformIntegral integrates a series whose values are "X per second" to estimate "total X so far"
+// if the series represents "X in this sampling interval" instead, then you should use transformCumulative.
+func transformIntegral(values []float64, parameters []value, timerange api.Timerange) ([]float64, error) {
```
achow: A thought about integral.. This behavior of scaling up each entry only makes sense if things get resampled using average instead of sum. Normally the `transformCumulative` would make sense because all data points in between are accounted for where the resolution boundaries are.

Nathan-Fenner: Yeah, I agree that the integral would not commonly be useful (if ever), but I think it would be confusing if doing `integral(derivative( series ))` gave you something different than you started with.

query/transformation.go
Nathan-Fenner: `transform`s now only take the resolution (although I called it `scale`... I think it will be in units of seconds even if resolution is in finer units)

query/transformation.go
Nathan-Fenner: The resolution of a series can't be chosen as an argument, it's computed from the properties of the given serieslist. So, from a caller's perspective, there's no way to select it.

`transformIntegral` makes sense for things that automatically keep track of rates in units of time (genuine rates in units of events/second or similar) 

`transormCumulative` makes sense for metrics that have counts like "messages received since last sample" which don't take the actual passage of time into account.

(On a related note, I just realized that a sampling method would have to be "sum" rather than "average" to deal with that last case - otherwise the resolution you'd sample at would totally change how the data looks).

jeeyoungk:
LGTM on my side!

PR #42, opened 2015-06-02
Make negative match on keywords stricter

achow:
@jeeyoungk @Nathan-Fenner


query/language.peg
```diff
@@ -192,7 +192,7 @@ METRIC_NAME <- IDENTIFIER
 TAG_NAME <-    IDENTIFIER
 # TODO - may be refactored later.
 TIMESTAMP <-   NUMBER_NATURAL / STRING
-IDENTIFIER <-  "`" CHAR* "`" / !KEYWORD ID_SEGMENT ("." ID_SEGMENT)*
+IDENTIFIER <-  "`" CHAR* "`" / !(KEYWORD !ID_SEGMENT) ID_SEGMENT ("." ID_SEGMENT)*
```
Nathan-Fenner: Won't this mean that an IDENTIFIER will match something that goes `KEYWORD ID_SEGMENT`?

Since `KEYWORD` would consume `KEYWORD`, and them `ID_SEGMENT` would match `ID_SEGMENT`, the match `(KEYWORD !ID_SEGMENT)` would fail so `!(KEYWORD !ID_SEGMENT)` would succeed, and then you'd match `ID_SEGMENT` on the `KEYWORD` at the beginning.

Nathan-Fenner: It should be `!ID_CONT` instead of `!ID_SEGMENT`, right?

Otherwise `"in3"` would be matched as a failing keyword, but `"inA"` would not.

achow: Good catch. Fixed.

Nathan-Fenner:
LGTM

jeeyoungk:
LGTM

jeeyoungk:
@achow need to rebase and rerun `build.sh`

PR #43, opened 2015-06-03
Refactor expressions to return Value instead of *api.SeriesList

Nathan-Fenner:
To better enable string and scalar parameters to functions, `Expression`s return a `Value` interface type instead of always returning `*api.SeriesList`. The `Value` interface defines methods to cast (with possible error returned) to `api.SeriesList`, `string`, or `float64`. There are three instances of the `Value` interface defined, one for each type. Each can only be casted to its own value, except for `ScalarValue` which can produce an `api.SeriesList`.

This change required modifying the corresponding test file to work with the new types.

@jeeyoungk @achow

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -138,10 +138,10 @@ type Timerange struct {
 
 // IsValid determines whether the given timerange meets the constraint.
 func (tr Timerange) IsValid() bool {
-	return (tr.Start%tr.Resolution == 0 &&
+	return tr.Resolution > 0 &&
+		tr.Start%tr.Resolution == 0 &&
 		tr.End%tr.Resolution == 0 &&
-		tr.Resolution > 0 &&
-		tr.Start <= tr.End)
+		tr.Start <= tr.End
 }
 
 // Slots represent the total # of data points
--- a/query/expression.go
+++ b/query/expression.go
@@ -51,7 +51,7 @@ type ConversionError struct {
 }
 
 func (e ConversionError) Error() string {
-	return fmt.Sprintf("cannot convert from type %from to type %to%")
+	return fmt.Sprintf("cannot convert from type %s to type %s", e.from, e.to)
 }
 
 // A SeriesListValue is a value which holds a SeriesList
--- a/query/expression.go
+++ b/query/expression.go
@@ -37,53 +37,53 @@ type EvaluationContext struct {
 	SampleMethod api.SampleMethod
 }
 
-// A Value is the result of evaluating an expression.
+// A value is the result of evaluating an expression.
 // They can be floating point values, strings, or series lists.
-type Value interface {
-	ToSeriesList(api.Timerange) (api.SeriesList, error)
-	ToString() (string, error)
-	ToScalar() (float64, error)
+type value interface {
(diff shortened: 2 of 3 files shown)


api/types.go
```diff
@@ -138,10 +138,10 @@ type Timerange struct {
...
 // IsValid determines whether the given timerange meets the constraint.
 func (tr Timerange) IsValid() bool {
-	return (tr.Start%tr.Resolution == 0 &&
+	return tr.Resolution > 0 &&
+		tr.Start%tr.Resolution == 0 &&
 		tr.End%tr.Resolution == 0 &&
-		tr.Resolution > 0 &&
-		tr.Start <= tr.End)
+		tr.Start <= tr.End
 }
```
Nathan-Fenner: I came across this when a panic happened while I was trying to fix up the tests to work with the new expression types.

Performing modulo by 0 causes a panic (like division by 0), so `tr.Resolution > 0` should be checked before the other cases.

jeeyoungk: +1. That's a good catch.

query/expression.go
```diff
@@ -37,6 +37,76 @@ type EvaluationContext struct {
...
+	from string
+	to   string
+}
+
+func (e ConversionError) Error() string {
+	return fmt.Sprintf("cannot convert from type %s to type %s", e.from, e.to)
+}
+
+// A SeriesListValue is a value which holds a SeriesList
+type SeriesListValue api.SeriesList
```
jeeyoungk: URGHK. Types are from different packages and cannot implement methods. Bummer.

query/expression.go
```diff
@@ -46,34 +116,34 @@ type EvaluationContext struct {
 // Expressions correspond to the timerange in the current EvaluationContext.
 type Expression interface {
 	// Evaluate the given expression.
-	Evaluate(context EvaluationContext) (*api.SeriesList, error)
+	Evaluate(context EvaluationContext) (Value, error)
+}
+
+func EvaluateToSeriesList(e Expression, context EvaluationContext) (api.SeriesList, error) {
```
jeeyoungk: nit: don't think this needs to be public?

Nathan-Fenner: Which part? Do any of `Expression`, `Evaluate`, `EvaluationContext`, or `Value` need to be public?

Nathan-Fenner: Since it only uses public types and public methods/fields, even if it wasn't exported, an outside package could write it themselves...

Maybe some of the types should be private, like `Value`?

Nathan-Fenner: I added a new commit where I unexported `value`, `seriesListValue`, `stringValue`, `scalarValue`, and `conversionError`.

jeeyoungk: in fact, we're only using it in test - so maybe move to tests?

jeeyoungk:
LGTM

query/expression.go
Nathan-Fenner: Although I haven't done it yet, I think I'm going to use it to wire up aggregation/transformation functions, so I'm going to leave it where it is for now

PR #44, opened 2015-06-03
Updating logging to use standard logging instead of glog instead.

jeeyoungk:
@achow @Nathan-Fenner

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -24,8 +24,8 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"log"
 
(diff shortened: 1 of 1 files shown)


Nathan-Fenner:
CI script says gofmt needs to be run on blueflood.go

achow:
The thing that sucks about removing glog is we lose the ability to filter by log level, so things might just get really verbose (and all go to stdout)... Maybe we should just comment it out and revisit later?

Otherwise LGTM after getting the build to pass.

jeeyoungk:
@achow yes. the level issue is real. :/

I will think about a better solution, meanwhile i think making integration between public & private square repo is better priority.

PR #45, opened 2015-06-03
String literal parsing improvements

Nathan-Fenner:
This PR changes several behaviors about string literal parsing:

* String literals now return their contents. Writing `STRING { fmt.printf("%s", buffer[start:end]) }` will now print `Hello!` when parsing the string `"Hello!"` or `'Hello!`' (rather than including the quotes in the output).
* Also, different quotes don't need to be escaped. `"These 'single quotes' don't need to be escaped"` will be parsed correctly (no backslashes needed).
* You'd still need to write `'These \'single quotes\' do need backslashes'` since they match. Also, unfortunately, the resulting string will contain the backslashes, which means that code may need to check for their presence/absence.

@jeeyoungk @achow

### DIFF
--- a/query/parser.go
+++ b/query/parser.go
@@ -70,12 +70,12 @@ import (
 // Time zones are ommitted where only the date is given (perhaps this should be changed)
 // TODO: use a less brittle method than putting the single quotes inside these
 var dateFormats = []string{
-	"'2006-1-2 15:04:05 MST'",
-	"'2006-1-2 15:04 MST'",
-	"'2006-1-2 15 MST'",
-	"'2006-1-2 MST'",
-	"'2006-1-2'",
-	"'2006-1'",
+	"2006-1-2 15:04:05 MST",
+	"2006-1-2 15:04 MST",
+	"2006-1-2 15 MST",
+	"2006-1-2 MST",
+	"2006-1-2",
+	"2006-1",
 }
 
 func parseDate(date string) (int64, error) {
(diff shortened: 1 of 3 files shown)


query/parser.go
```diff
@@ -68,14 +68,35 @@ import (
...
+	"2 Jan 2006 15:04:05 MST",
+	"2 Jan 2006 15:04 MST",
+	"2 Jan 2006 15 MST",
+	"2 Jan 2006 MST",
+	"2 Jan 2006",
+	time.ANSIC,
+	time.UnixDate,
+	time.RubyDate,
+	time.RFC822,
+	time.RFC822Z,
```
Nathan-Fenner: I needed to update the dates to get rid of the single-quotes they previously contained, so while I was at it I added more date formats.

(These would be the acceptable formats for timestamps in `from` and `to` property clauses).

achow:
Needs a rebase against master.

We may want to do something with identifiers so that behavior is consistent and remove the backtick stripping from `unescapeLiteral()`.

Personally, I'd want the rules that are capturing but don't have an associated block to be tagged as such, but otherwise LGTM.

jeeyoungk:
@achow @Nathan-Fenner 

I think the most "correct" way of doing this is actually moving the string building to the parser side, instead of `unescapeLiteral()`.

I.e. make the parser into a state machine that goes. "whenever non-escaped character is encountered, add it to buffer. whenever a escaped character is encountered, add the escaped version to the buffer. at the end, convert the buffer to string".

I was lazy (just like everything else I do), and converting the entire raw string into a s sanitized string - which is convenient but has weird edge cases like the one you've described.

jeeyoungk:
LGTM

PR #47, opened 2015-06-03
Expression function implementation for aggregates

Nathan-Fenner:
This PR implements calls to the 4 implemented aggregation functions.

It also performs ancillary cleanup for arithmetic calls.

### DIFF
--- a/query/aggregate/aggregate.go
+++ b/query/aggregate/aggregate.go
@@ -72,7 +72,7 @@ func groupBy(list api.SeriesList, tags []string) []group {
 	return result
 }
 
-var AggregateMap = map[string]func([]float64) float64{
+var aggregateMap = map[string]func([]float64) float64{
 	"aggregate.sum":
 	// The aggregatefor sum finds the sum of the given array.
 	func(array []float64) float64 {
@@ -121,6 +121,17 @@ var AggregateMap = map[string]func([]float64) float64{
 		}
 		return max
 	},
+}
+
+// GetAggregate gives the aggregate of the given name (and false if it doesn't exist).
+func GetAggregate(name string) (func([]float64) float64, bool) {
+	aggregate, ok := aggregateMap[name]
+	return aggregate, ok
+}
+
+// AddAggregate adds an aggregate of a given name to the aggregate map.
+func NewAggregate(name string, aggregate func([]float64) float64) {
+	aggregateMap[name] = aggregate
 }
 
 // applyAggregation takes an aggregation function ( [float64] => float64 ) and applies it to a given list of Timeseries
--- a/query/aggregate/aggregate.go
+++ b/query/aggregate/aggregate.go
@@ -18,8 +18,6 @@ package aggregate
 // and produces an aggregated SeriesList with one list per group, each group having been aggregated into it.
 
 import (
-	"errors"
-	"fmt"
 	"math"
 
 	"github.com/square/metrics/api"
@@ -74,7 +72,7 @@ func groupBy(list api.SeriesList, tags []string) []group {
 	return result
 }
 
-var aggregateMap = map[string]func([]float64) float64{
+var AggregateMap = map[string]func([]float64) float64{
 	"aggregate.sum":
 	// The aggregatefor sum finds the sum of the given array.
 	func(array []float64) float64 {
@@ -162,11 +160,7 @@ func applyAggregation(group group, aggregator func([]float64) float64) api.Times
 // `aggregateBy` takes a series list, an aggregator, and a set of tags.
 // It produces a SeriesList which is the result of grouping by the tags and then aggregating each group
 // into a single Series.
-func AggregateBy(list api.SeriesList, name string, tags []string) (api.SeriesList, error) {
-	aggregator, ok := aggregateMap[name]
-	if !ok {
-		return api.SeriesList{}, errors.New(fmt.Sprintf("no such aggregation function '%s'", name))
-	}
+func AggregateBy(list api.SeriesList, aggregator func([]float64) float64, tags []string) api.SeriesList {
 	// Begin by grouping the input:
 	groups := groupBy(list, tags)
 
@@ -180,5 +174,5 @@ func AggregateBy(list api.SeriesList, name string, tags []string) (api.SeriesLis
 		// The group contains a list of Series and a TagSet.
 		result.Series[i] = applyAggregation(group, aggregator)
 	}
-	return result, nil
+	return result
 }
--- a/query/aggregate/aggregate.go
+++ b/query/aggregate/aggregate.go
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package query
+package aggregate
 
 // This file culminates in the definition of `aggregateBy`, which takes a SeriesList and an Aggregator and a list of tags,
 // and produces an aggregated SeriesList with one list per group, each group having been aggregated into it.
 
 import (
+	"errors"
+	"fmt"
 	"math"
 
 	"github.com/square/metrics/api"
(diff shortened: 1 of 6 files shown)


query/expression.go
```diff
@@ -144,19 +144,66 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
...
-	case "/":
-		return evaluateBinaryOperation(context, expr.functionName, expr.arguments,
-			func(left, right float64) float64 { return left / right })
+		fallthrough
 	case "*":
-		return evaluateBinaryOperation(context, expr.functionName, expr.arguments,
-			func(left, right float64) float64 { return left * right })
+		fallthrough
+	case "/":
+		arguments := make([]value, len(expr.arguments))
```
Nathan-Fenner: Although it's possible to move this definition up to the top of the structure (especially since it's repeated for the aggregate functions), in order to change the `Timerange` to accommodate time shifting, arguments can't be evaluated until the function name is known (and then they may need to be evaluated in a special way).

query/expression.go
```diff
@@ -144,19 +144,66 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
...
+			"aggregate.max":  maxAggregate,
+		}
+		if len(arguments) != 1 {
+			return nil, errors.New(fmt.Sprintf("Function `%s` expected 2 operands but received %d (%+v)", name, len(arguments), arguments))
+		}
+		list, err := arguments[0].toSeriesList(context.Timerange)
+		if err != nil {
+			return nil, err
+		}
+		return seriesListValue(aggregateBy(list, aggregateMap[funMap[name]], expr.groupBy)), nil
```
achow: `aggregateMap` here is really confusing since it appears to come out of nowhere. The `*Aggregate` enums also do the same... Maybe we should move these function implementations to their own package or something? How things are set up makes some of these references a little too magic for my liking, though maybe that's just how go works.

@jeeyoungk

Nathan-Fenner: one solution might be to replace the `aggregateMap` in `aggregate.go` (which is of type `map[aggregate]func([]float64) float64`) into the map defined here directly, giving it the new type `map[string]func([]float64) float64` and eliminating the `Aggregate` enum...

There needs to be some mapping from `String` to `func([]float64) float64` somewhere, though it wouldn't need to necessarily be a map (it could also be a list of structs or something similar, each a name-function pair).

achow: Imo, callers should only be concerned with the enum, and the callees should do the mappings from enums to implementation. I wonder if there's a better way in general to do dispatch though.

Nathan-Fenner: There was some previous discussion: https://github.com/square/metrics/pull/35#discussion_r31552407

What should be done here? Currently there is a "double" lookup, converting the `string` to the `Aggregate` through the local-variable `funMap`, and then again through `aggregateMap` (defined in `aggregate.go`).

The lookup from `enum => aggregate` could be done inside the `aggregateBy` function instead.

achow: Yes, have the enum => aggregate mapping occur in `aggregateBy`.

I still think we should export the enums and `aggregateBy`, not for package-external visibility, but for package-internal identification of things that are meant to be used. Use of unexported things should be limited to the files in which they are defined. Though, if we start doing this it _may_ be useful to start splitting these into packages.

Nathan-Fenner: How should we convert the function name (string) into the associated enum?

Should aggregate just define a `map[string]aggregate` and callers look up their given function name before calling `aggregateBy`?

Nathan-Fenner: I exported `AggregateBy` and also `AggregateMap` but I'm not sure about that one, and also made `AggregateBy` do the lookup (`aggregate => func`) itself.

Nathan-Fenner:
I moved `aggregate.go` and `aggregate_test.go` into a new package `query/aggregate`.

In addition, I changed the interface of `AggregateBy` from:

`func AggregateBy(list api.SeriesList, aggregator aggregate, tags []string) api.SeriesList`

to:

`func AggregateBy(list api.SeriesList, name string, tags []string) (api.SeriesList, error)`

The principle change is that an enum is no longer used to decide which aggregator is run- instead, a string is used.

The reason for this change is that the use of an enum provides little safety over pure string keys, the reason for this being that AggregateBy is *always* called only using user-supplied data (the function name) and never by the code itself, where verifying that the enum actually exists is possible.

Since users always supplied the data in the form of a string, with enums, it was necessary to perform two transformations: (`string => enum type`) and then (`enum type => function`). However, since the intermediate result is never used, except to be immediately transformed into the function by looking it up in the map, the intermediate result is not particularly useful.

`AggregateBy` now instead performs the lookup itself using the string, which means only one table of `name` / `func` needs to be defined (it's currently the unexported name `aggregateMap`). If the given name is not present in the table, then `AggregateBy` returns an error.

query/expression.go
```diff
@@ -144,19 +145,52 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
...
-	case "/":
-		return evaluateBinaryOperation(context, expr.functionName, expr.arguments,
-			func(left, right float64) float64 { return left / right })
-	case "*":
-		return evaluateBinaryOperation(context, expr.functionName, expr.arguments,
-			func(left, right float64) float64 { return left * right })
+
+	name := expr.functionName
+	switch name {
+	case "+", "-", "*", "/":
```
jeeyoungk: we should pull out `operatorMap` out, and have if statement on whether the value belongs on the `operatorMap` - this ensure that we don't have to manually keep `case` and `operatorMap` in sync.

query/expression.go
```diff
@@ -144,19 +145,52 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
...
+		left := arguments[0]
+		right := arguments[1]
+		operatorMap := map[string]func(float64, float64) float64{
+			"+": func(x, y float64) float64 { return x + y },
+			"-": func(x, y float64) float64 { return x - y },
+			"*": func(x, y float64) float64 { return x * y },
+			"/": func(x, y float64) float64 { return x / y },
+		}
+		return evaluateBinaryOperation(context, name, left, right, operatorMap[name])
+	case "aggregate.sum", "aggregate.mean", "aggregate.min", "aggregate.max":
```
jeeyoungk: same as here - i think the `case` statement should use values from some map.

query/aggregate/aggregate.go
```diff
@@ -72,17 +74,8 @@ func groupBy(list api.SeriesList, tags []string) []group {
...
-const (
-	sumAggregate aggregate = iota + 1
-	meanAggregate
-	minAggregate
-	maxAggregate
-)
-
-var aggregateMap = map[aggregate]func([]float64) float64{
-	sumAggregate:
+var aggregateMap = map[string]func([]float64) float64{
```
jeeyoungk: we can make this code more "modular" by instead of hard-coding `map[name]function`, expose a function called `NewAggregate(name string, func function)` which can be used to add functions. In that case, you can define these aggregate functions in different party, and you can compose them that way.

query/expression.go
Nathan-Fenner: The `switch`/`case` has been replaced by `value, ok :=` lookups in hashmaps (through a function `GetAggregate` for aggregation functions)

query/aggregate/aggregate.go
Nathan-Fenner: `aggregateMap` is now accessed through `GetAggregate` and `NewAggregate` functions; `GetAggregate` returns `func([]float64) float64, bool` for `ok` checks, and `NewAggregate` takes a name and a function.

query/aggregate/aggregate.go
```diff
@@ -133,6 +123,17 @@ var aggregateMap = map[aggregate]func([]float64) float64{
...
 
+// GetAggregate gives the aggregate of the given name (and false if it doesn't exist).
+func GetAggregate(name string) (func([]float64) float64, bool) {
+	aggregate, ok := aggregateMap[name]
+	return aggregate, ok
+}
+
+// AddAggregate adds an aggregate of a given name to the aggregate map.
+func NewAggregate(name string, aggregate func([]float64) float64) {
+	aggregateMap[name] = aggregate
```
jeeyoungk: let's `panic` on a duplicate registration.

jeeyoungk:
LGTM with nit

PR #48, opened 2015-06-04
Fix metric fetches

achow:
@jeeyoungk @Nathan-Fenner

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -72,11 +72,50 @@ func (b *Blueflood) Api() api.API {
 }
 
 func (b *Blueflood) FetchSeries(metric api.TaggedMetric, predicate api.Predicate, sampleMethod api.SampleMethod, timerange api.Timerange) (*api.SeriesList, error) {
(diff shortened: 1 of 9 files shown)


Nathan-Fenner:
LGTM

api/backend/blueflood/blueflood.go
```diff
@@ -72,11 +72,50 @@ func (b *Blueflood) Api() api.API {
...
-	graphiteMetric, err := b.api.ToGraphiteName(metric)
+	metricTagSets, err := b.Api().GetAllTags(metric.MetricKey)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: Be a little smarter about this?
+	resultSeries := []api.Timeseries{}
+	for _, ts := range metricTagSets {
+		if predicate == nil || predicate.Apply(ts) {
```
jeeyoungk: i think we should pass around all-matching predicate rather than `nil` predicate. Empty object pattern :P

achow: Well, either way works. All-matching predicates would match the `Apply`, and nil is technically a valid argument. If we want to explicitly disallow nil predicates, we can.

PR #50, opened 2015-06-08
Expressions using Transformations

Nathan-Fenner:
Through the usage of `GetTransformation`, expression is now capable of performing transformation invocations.

### DIFF
--- a/query/expression.go
+++ b/query/expression.go
@@ -201,6 +201,35 @@ func (expr *functionExpression) Evaluate(context EvaluationContext) (value, erro
 		return seriesListValue(series), nil
 	}
 
+	if transform, ok := GetTransformation(name); ok {
+		//Verify that at least one argument is given.
+		if len(expr.arguments) != 1 {
+			return nil, errors.New(fmt.Sprintf("Function `%s` expects at least 1 argument but was given 0", name))
+		}
+		first, err := expr.arguments[0].Evaluate(context)
+		if err != nil {
+			return nil, err
+		}
+		list, err := first.toSeriesList(context.Timerange)
+		if err != nil {
+			return nil, err
+		}
+		// Evaluate all the other parameters:
+		rest := expr.arguments[1:]
+		parameters := make([]value, len(rest))
+		for i := range parameters {
+			parameters[i], err = rest[i].Evaluate(context)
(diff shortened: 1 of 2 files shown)


jeeyoungk:
LGTM

query/expression.go
```diff
@@ -201,6 +201,35 @@ func (expr *functionExpression) Evaluate(context EvaluationContext) (value, erro
...
+		first, err := expr.arguments[0].Evaluate(context)
+		if err != nil {
+			return nil, err
+		}
+		list, err := first.toSeriesList(context.Timerange)
+		if err != nil {
+			return nil, err
+		}
+		// Evaluate all the other parameters:
+		rest := expr.arguments[1:]
```
achow: Isn't this dead code since we enforce `len(expr.arguments) == 1`?

Nathan-Fenner: Aggregates have the requirement that there is *only* one argument; transformations require *at least* one argument. Individual transformations each check how many more are expected. For example, moving average has a second one for the time over which to average, while derivative wants only the series.

query/expression.go
```diff
@@ -201,6 +201,35 @@ func (expr *functionExpression) Evaluate(context EvaluationContext) (value, erro
 		return seriesListValue(series), nil
 	}
 
+	if transform, ok := GetTransformation(name); ok {
+		//Verify that at least one argument is given.
+		if len(expr.arguments) != 1 {
```
achow: Looks like you need to fix this then.

Nathan-Fenner: Ah, yes

query/transformation.go
```diff
@@ -193,3 +193,24 @@ func transformMapMaker(name string, fun func(float64) float64) func([]float64, [
...
+}
+
+func GetTransformation(name string) (transform, bool) {
+	transform, ok := transformTable[name]
+	return transform, ok
+}
+
+func NewTransformation(name string, transform transform) {
+	if _, ok := transformTable[name]; ok {
+		panic(fmt.Sprintf("transformation `%s` has already been declared", name))
```
achow: We shouldn't be using panics explicitly... This is easily a recoverable error.

panics should be reserved for catastrophic exceptions.

PR #51, opened 2015-06-08
Add timeshift function

Nathan-Fenner:
This PR adds a special case for `timeshift` as a function to be invoked by expressions. It works by modifying the evaluation context before evaluating its first argument.

Currently, timeshifting is a special case, since no other function or operation during expression evaluation actually changes the `EvaluationContext`.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -157,13 +157,22 @@ func (tr Timerange) IsValid() bool {
 		tr.Start <= tr.End
 }
 
+func rounder(x, r int64) int64 {
+	if x < 0 {
+		return -rounder(-x, r)
+	}
+	// This performs a round
+	return (x + r/2) / r * r
+}
+
 // Round() will fix some invalid timeranges by rounding their starts and ends.
 func (tr Timerange) Round() Timerange {
+
 	if tr.Resolution == 0 {
 		return tr
 	}
-	tr.Start = (tr.Start + tr.Resolution/2) / tr.Resolution * tr.Resolution // This performs a round
-	tr.End = (tr.End + tr.Resolution/2) / tr.Resolution * tr.Resolution     // This performs a round
+	tr.Start = rounder(tr.Start, tr.Resolution)
+	tr.End = rounder(tr.End, tr.Resolution)
 	return tr
 }
 
--- a/api/types.go
+++ b/api/types.go
@@ -161,7 +161,14 @@ func snap(n, boundary int64) int64 {
 	if n < 0 {
 		return -snap(-n, boundary)
 	}
-	// This performs a round
+	// This performs a round.
+	// Dividing by `boundary` truncates towards zero.
+	// The resulting integer is then multiplied by `boundary` again.
+	// Thus the result is a multiple of `boundary`.
+	// For integer division, x/r*r = (x/r)*r in general rounds to a multiple of r towards 0.
+	// Adding `boundary/2` changes this instead to a "round to nearest" rather than "round towards 0".
+	// (Where "up" is the round for values exactly halfway between).
+	// These halfway points round "away from zero" (rather than "towards -infinity").
 	return (n + boundary/2) / boundary * boundary
 }
 
--- a/api/types.go
+++ b/api/types.go
@@ -157,6 +157,23 @@ func (tr Timerange) IsValid() bool {
 		tr.Start <= tr.End
 }
 
+// Round() will fix some invalid timeranges by rounding their starts and ends.
+func (tr Timerange) Round() Timerange {
+	if tr.Resolution == 0 {
+		return tr
+	}
+	tr.Start = (tr.Start + tr.Resolution/2) / tr.Resolution * tr.Resolution // This performs a round
+	tr.End = (tr.End + tr.Resolution/2) / tr.Resolution * tr.Resolution     // This performs a round
+	return tr
+}
+
+// Later() returns a timerange which is forward in time by the amount given
+func (tr Timerange) Later(time int64) Timerange {
+	tr.Start += time
+	tr.End += time
+	return tr.Round()
+}
+
 // Slots represent the total # of data points
 // Behavior is undefined when operating on an invalid Timerange. There's a
 // circular dependency here, but it all works out.
--- a/api/types.go
+++ b/api/types.go
@@ -157,30 +157,30 @@ func (tr Timerange) IsValid() bool {
 		tr.Start <= tr.End
 }
 
-func rounder(x, r int64) int64 {
-	if x < 0 {
-		return -rounder(-x, r)
+func snap(n, boundary int64) int64 {
+	if n < 0 {
+		return -snap(-n, boundary)
 	}
 	// This performs a round
-	return (x + r/2) / r * r
+	return (n + boundary/2) / boundary * boundary
 }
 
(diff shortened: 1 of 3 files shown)


query/expression.go
```diff
@@ -109,6 +110,21 @@ func (value scalarValue) toScalar() (float64, error) {
 	return float64(value), nil
 }
 
+// toDuration will take a value, convert it to a string, and then parse it.
+// It produces a nanosecond count as a signed int64.
+// the valid ns, us (µs), ms, s, m, h
+func toDuration(value value) (int64, error) {
```
jeeyoungk: let's return `time.Duration` - i think it's not bad.

query/expression.go
```diff
@@ -230,6 +246,26 @@ func (expr *functionExpression) Evaluate(context EvaluationContext) (value, erro
...
+		if err != nil {
+			return nil, err
+		}
+		durationNano, err := toDuration(shift)
+		if err != nil {
+			return nil, err
+		}
+		newContext := context
+		newContext.Timerange = newContext.Timerange.Later(durationNano)
+		return expr.arguments[0].Evaluate(newContext)
```
jeeyoungk: we would want to rewrite the timerange of the response. to be the current timestamp, not the shifted timestamp.

api/types.go
```diff
@@ -157,6 +157,23 @@ func (tr Timerange) IsValid() bool {
...
+	if tr.Resolution == 0 {
+		return tr
+	}
+	tr.Start = (tr.Start + tr.Resolution/2) / tr.Resolution * tr.Resolution // This performs a round
+	tr.End = (tr.End + tr.Resolution/2) / tr.Resolution * tr.Resolution     // This performs a round
+	return tr
+}
+
+// Later() returns a timerange which is forward in time by the amount given
+func (tr Timerange) Later(time int64) Timerange {
```
jeeyoungk: is `x.Start() - x.End()` guaranteed to have `y.Start() - y.End()`, where `y = x.Later(range)` for any range? I think the answer is yes, but I want to be confident that is the property we have (since time-shifted fetch should hopefully fetch the same amount of data)

Nathan-Fenner: Yes, that's important to maintain. In terms of a proof of correctness or similar, if you consider:

`start` and `end` are the two variables, both multiples of `resolution` (short `res`) since the input must be well-formed, so let `start = res*s` and `end = res*e`, so our rounding ends up with `(res*s + shift + res/2) / res * res` for the new start, and `(res*e + shift + res/2)/res*res` for our new end (all divisions are integer division).

`(res*s + shift + res/2)/res*res = (res*s)/res*res + (shift + res/2)/res*res` since `(res*s)` is an exact multiple of `res`, so we can distribute the division, and then `= (s*res)/res*res + (shift + res/2)/res*res = s*res + (shift + res/2)/res`.

Similarly, the new end will be `e*res + (shift + res/2)/res`. So their differences will still be `e*res - s*res` which is exactly `end - start`.

Having said that, test cases would be good and I can add them to api.types to verify this behavior.

Nathan-Fenner: So the above relies on the positive-ness of some of these properties so in order to fix that I made a new rounder function which is pretty similar; basically it rounds negative values correctly, and positive values the same.

query/expression.go
Nathan-Fenner: The timerange is now rewritten in the response.

api/types.go
achow: We shouldn't be assuming `start % resolution == 0`...

query/expression.go
achow: Why nanosecond?

api/types.go
Nathan-Fenner: Currently, this is enforced in the `.IsValid()` method for Timeranges. If it's not enforced then `.Later()` becomes much, much simpler.

jeeyoungk: @achow on the assumption of `start % resolution == 0` - I had it because it seemed like a convenient restriction to have.

query/expression.go
Nathan-Fenner: The formats are just those supported by `time.ParseDuration()` which represents time with nanosecond resolution... Currently I wasn't actually scaling at all, which would probably behave really strangely...

The comments for `api.Timerange` suggest that it's currently supposed to represent seconds, but will later represent milliseconds?

Has this been properly established yet?

api/types.go
```diff
@@ -157,6 +157,32 @@ func (tr Timerange) IsValid() bool {
 		tr.Start <= tr.End
 }
 
+func rounder(x, r int64) int64 {
+	if x < 0 {
+		return -rounder(-x, r)
+	}
+	// This performs a round
+	return (x + r/2) / r * r
```
achow: `x / r * r = r`?

api/types.go
achow: I'd name this `Shift`. `Later` sounds a bit weird and isn't quite how I'd expect to shift time ranges.

api/types.go
```diff
@@ -157,6 +157,32 @@ func (tr Timerange) IsValid() bool {
 		tr.Start <= tr.End
 }
 
+func rounder(x, r int64) int64 {
```
achow: I'd name this `snap(n, boundary)`.

api/types.go
```diff
@@ -157,6 +157,32 @@ func (tr Timerange) IsValid() bool {
...
+	if x < 0 {
+		return -rounder(-x, r)
+	}
+	// This performs a round
+	return (x + r/2) / r * r
+}
+
+// Round() will fix some invalid timeranges by rounding their starts and ends.
+func (tr Timerange) Round() Timerange {
```
achow: I'd name this `Snap` as well.

achow:
High level note: maybe we should unexport `Timerange` fields and create them using `NewTimerange` if we want to enforce validity. Requiring use of `IsValid()` to check and delegating to callers to enforce validity/assumptions is a bit ugly...

@jeeyoungk

api/types.go
Nathan-Fenner: Since `x / r * r = (x / r) * r`, and `x / r` truncates towards zero, this accomplishes a round (towards 0). A round "towards nearest" (with halfways rounding up) is accomplished by the `+ r/2`.

Unfortunately, the golang `math` (or any other I know of) library does not provide an equivalent...

It's definitely not the most clear expression...

jeeyoungk:
@achow yes... I've been thinking about that too. have `NewTimerange(fields) (Timerange, err)` - this will ensure that we will only have valid time ranges. I think it's a good refactoring to do.

api/types.go
achow: You can probably `math.Floor(float64(x) + float64(r) / 2)`?

Nathan-Fenner: Some extremely large numbers aren't exactly representable as `float64` (2^53 or about around 10^16) so I'm a little concerned about introducing floating point error. In practice I wouldn't expect it to be a concern, though.

achow: Should leave a comment explaining what you're trying to do here. It doesn't have to be too detailed, but at first glance the `/ r * r` isn't quite clear.

Nathan-Fenner: That's a good idea. I added documentary comments explaining how it works.

jeeyoungk:
LGTM

PR #53, opened 2015-06-08
Use errors instead of panics for NewAggregation/NewTransformation

Nathan-Fenner:


achow:
I'd go with something like `RegisterAggregate` instead of `NewAggregate`. `New*` is more for factories that give you a built struct. Otherwise LGTM.

jeeyoungk:
@achow yes on naming convention!

jeeyoungk:
LGTM

PR #54, opened 2015-06-08
Unexport fields from api.Timerange and Standardize Milliseconds

Nathan-Fenner:
This commit un-exports the fields `start`, `end`, and `resolution` from `api.Timerange`. To read them, the methods `.Start()`, `.End()`, and `.Resolution()` are provided.

To construct a new timerange, invoke `api.NewTimerange(start, end, resolution)` which returns `(api.Timerange, error)`.

All instances of `api.Timerange{ ... }` have been replaced with a call to `api.NewTimerange(...)` or `api.DefaultTimerange()` so that all instantiated timeranges are valid, even if they're never used.

To create a valid default timerange, you can call `api.DefaultTimerange()` which gives a timerange corresponding to `{0, 0, 30}`. 

It's still possible to construct an invalid timerange, namely the "zero" timerange `api.Timerange{}`. This behavior is nowhere used (they have all been replaced with `api.DefaultTimerange()`).

@jeeyoungk @achow

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -140,8 +140,8 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 	}
 
 	params := url.Values{}
-	params.Set("from", strconv.FormatInt(timerange.Start()*1000, 10))
-	params.Set("to", strconv.FormatInt(timerange.End()*1000, 10))
+	params.Set("from", strconv.FormatInt(timerange.Start(), 10))
+	params.Set("to", strconv.FormatInt(timerange.End(), 10))
 	params.Set("resolution", bluefloodResolution(timerange.Resolution()))
 	params.Set("select", fmt.Sprintf("numPoints,%s", strings.ToLower(selectResultField)))
 
@@ -184,15 +184,15 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 // between them.
 func bluefloodResolution(r int64) string {
 	switch {
-	case r < 5*60:
+	case r < 5*60*1000:
 		return ResolutionFull
-	case r < 20*60:
+	case r < 20*60*1000:
 		return Resolution5Min
-	case r < 60*60:
+	case r < 60*60*1000:
 		return Resolution20Min
-	case r < 240*60:
+	case r < 240*60*1000:
 		return Resolution60Min
-	case r < 1440*60:
+	case r < 1440*60*1000:
 		return Resolution240Min
 	}
 	return Resolution1440Min
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -140,9 +140,9 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 	}
 
 	params := url.Values{}
-	params.Set("from", strconv.FormatInt(timerange.Start*1000, 10))
-	params.Set("to", strconv.FormatInt(timerange.End*1000, 10))
-	params.Set("resolution", bluefloodResolution(timerange.Resolution))
+	params.Set("from", strconv.FormatInt(timerange.Start()*1000, 10))
+	params.Set("to", strconv.FormatInt(timerange.End()*1000, 10))
+	params.Set("resolution", bluefloodResolution(timerange.Resolution()))
 	params.Set("select", fmt.Sprintf("numPoints,%s", strings.ToLower(selectResultField)))
 
 	queryUrl.RawQuery = params.Encode()
--- a/api/types.go
+++ b/api/types.go
@@ -5,6 +5,7 @@ package api
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"regexp"
 	"sort"
 )
@@ -172,8 +173,17 @@ func (tr Timerange) Resolution() int64 {
 
 // NewTimerange creates a timerange which is validated, providing error otherwise.
 func NewTimerange(start, end, resolution int64) (*Timerange, error) {
-	if resolution == 0 || start%resolution != 0 || end%resolution != 0 || start > end {
-		return nil, errors.New("invalid timernage")
+	if resolution <= 0 {
+		return nil, errors.New(fmt.Sprintf("resolution must be more than 0 (resolution=%d)", resolution))
+	}
+	if start%resolution != 0 {
+		return nil, errors.New(fmt.Sprintf("start % resolution must be 0 (start=%d, resolution=%d)", start, resolution))
+	}
+	if end%resolution != 0 {
+		return nil, errors.new(fmt.Sprintf("end % resolution must be 0 (end=%d, resolution=%d)", end, resolution))
+	}
+	if start > end {
+		return nil, errors.new(fmt.Sprintf("start must be <= end (start=%d, end=%d)", start, end))
 	}
 	return &Timerange{start: start, end: end, resolution: resolution}, nil
 }
--- a/api/types.go
+++ b/api/types.go
@@ -170,20 +170,12 @@ func (tr Timerange) Resolution() int64 {
 	return tr.start
 }
 
-func NewTimerange(start, end, resolution int64) (Timerange, error) {
-	result := Timerange{start: start, end: end, resolution: resolution}
-	if !result.isValid() {
-		return Timerange{}, errors.New("invalid timerange")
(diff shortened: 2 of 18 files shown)


api/types.go
```diff
@@ -141,27 +142,55 @@ type SeriesType string
 // end =   0 mod resolution
 //
 // This range is inclusive of Start and End (i.e. [Start, End]). Start and End
-// are Unix second timestamps. Resolution is in seconds.
-// TODO: Make these ms
+// are Unix milliseconds timestamps. Resolution is in milliseconds.
+// (Millisecond precision allows an effective range of 290 million years in each direction)
```
achow: Probably unnecessary to note that millis range gives us space to last until our robot overloads take over and life ends :P

api/types.go
```diff
@@ -141,27 +142,55 @@ type SeriesType string
...
-	Start      int64
-	End        int64
-	Resolution int64
+	start      int64
+	end        int64
+	resolution int64
+}
+
+// DefaultTimerange is a valid timerange which can be safely used
+func DefaultTimerange() Timerange {
```
achow: Why?

api/types.go
```diff
@@ -141,27 +142,55 @@ type SeriesType string
...
+	return result, nil
 }
 
 // IsValid determines whether the given timerange meets the constraint.
-func (tr Timerange) IsValid() bool {
-	return tr.Resolution > 0 &&
-		tr.Start%tr.Resolution == 0 &&
-		tr.End%tr.Resolution == 0 &&
-		tr.Start <= tr.End
+func (tr Timerange) isValid() bool {
```
achow: Should inline this in the constructor so we can have better error messaging for all the things that can go wrong.

api/types.go
```diff
@@ -141,27 +142,55 @@ type SeriesType string
...
+func (tr Timerange) End() int64 {
+	return tr.start
+}
+
+// Resolution() returns the .resolution field
+func (tr Timerange) Resolution() int64 {
+	return tr.start
+}
+
+func NewTimerange(start, end, resolution int64) (Timerange, error) {
```
achow: Should return `*Timerange`

query/expression_test.go
```diff
@@ -58,16 +58,15 @@ func (expr *LiteralSeriesExpression) Evaluate(context EvaluationContext) (value,
 }
 
 func Test_ScalarExpression(t *testing.T) {
+	timerangeA, _ := api.NewTimerange(0, 10, 2)
```
achow: Should assert that this succeeds.

api/types.go
Nathan-Fenner: The original reason is that the timerange `{0, 0, 0}` is invalid, so any code that wanted to build a valid (but placeholder/useless) `Timerange` would be able to do so without having to check errors or discard them.

However, I did just realize that there are some issues elsewhere that need to be addressed with this change, namely in the contextevaluation generation.

api/types_test.go
```diff
@@ -123,11 +123,11 @@ func TestTimerange(t *testing.T) {
...
 		a := assert.New(t).Contextf("input=%d:%d:%d",
-			suite.Timerange.Start,
-			suite.Timerange.End,
-			suite.Timerange.Resolution,
+			suite.Timerange.start,
+			suite.Timerange.end,
+			suite.Timerange.resolution,
 		)
-		a.EqBool(suite.Timerange.IsValid(), suite.ExpectedValid)
+		a.EqBool(suite.Timerange.isValid(), suite.ExpectedValid)
```
achow: These tests should be converted to use the constructor instead of creating `Timeranges` on their own.

query/transformation.go
```diff
@@ -42,7 +42,7 @@ func transformTimeseries(series api.Timeseries, transform transform, parameters
 // timerangeScale takes an api.Timerange and computes the "scale" or duration of one sample (in seconds).
 // It is useful for transformations that normalize on time (like derivative or integral).
 func timerangeScale(timerange api.Timerange) float64 {
-	return float64(timerange.Resolution)
+	return float64(timerange.Resolution())
```
achow: Perhaps out of scope of the current changes, but this is unnecessary indirection.

api/types.go
achow: `Timerange{}` is fine if you really need a `Timerange` as a placeholder.

query/transformation.go
Nathan-Fenner: The `timerangeScale` function will convert to seconds (rather than milliseconds) once I make those changes everywhere (time units are still inconsistent for transformations/expressions I think) so there will be a division also.

achow: Any uses of `Timerange` fields should always expect millis. If it's necessary to convert between seconds and millis, that should be done in the context of whatever needs it (e.g. if your backend could only handle seconds, convert to seconds only when you're calling out).

Nathan-Fenner: This was the intent of the `timerangeScale` function. Transformation functions like `derivative` treat rates as though they are per-second (rather than per-millisecond) and so are given a "scale" whose units are seconds.

The intent was that series would then take on values like "3 per second" (stored directly as `float64(3)`) instead of "0.003 per millisecond". I'm not entirely sure how this behavior will interact with the actual visualization/reporting of the data.

achow: Derivative and such will still work if you keep all units as millis instead of seconds. If the final result needs to be in seconds, that's fine, but we should only do the conversion before sending back the result.

Nathan-Fenner: Now using milliseconds for natural units; as such the `timerangeScale` function has been removed.

api/types.go
```diff
@@ -141,20 +142,40 @@ type SeriesType string
...
+}
+
+// Resolution() returns the .resolution field
+func (tr Timerange) Resolution() int64 {
+	return tr.resolution
+}
+
+// NewTimerange creates a timerange which is validated, providing error otherwise.
+func NewTimerange(start, end, resolution int64) (*Timerange, error) {
+	if resolution == 0 || start%resolution != 0 || end%resolution != 0 || start > end {
```
achow: Should probably have different messages for each to make it clear what's wrong.

query/parser.go
```diff
@@ -108,7 +108,9 @@ func parseDate(date string) (int64, error) {
 	for _, format := range dateFormats {
 		t, err := time.Parse(format, date)
 		if err == nil {
-			return t.Unix(), nil
+			// Multiply the number of seconds by 1000 to convert to milliseconds,
```
achow: You don't have to be so specific with the comments. Something like `Convert to millis` or just a top-level comment on `parseDate` saying return value is in timestamp millis would be sufficient (preferably latter).

It's a bit of an art to figure out where to comment and how much to, so it might take a while to get the hang of.

api/types.go
```diff
@@ -141,20 +143,49 @@ type SeriesType string
...
+	return tr.end
+}
+
+// Resolution() returns the .resolution field
+func (tr Timerange) Resolution() int64 {
+	return tr.resolution
+}
+
+// NewTimerange creates a timerange which is validated, providing error otherwise.
+func NewTimerange(start, end, resolution int64) (*Timerange, error) {
```
jeeyoungk: I think it's unidiomatic go to give pointers to simple structs - probably less efficient too.

that said, the topic of whether we should return pointers or not in many different places is a highly debatable topic - i'm still not sure which case should be which.

api/backend/blueflood/blueflood.go
```diff
@@ -140,9 +140,9 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 	}
 
 	params := url.Values{}
-	params.Set("from", strconv.FormatInt(timerange.Start*1000, 10))
-	params.Set("to", strconv.FormatInt(timerange.End*1000, 10))
-	params.Set("resolution", bluefloodResolution(timerange.Resolution))
+	params.Set("from", strconv.FormatInt(timerange.Start(), 10))
+	params.Set("to", strconv.FormatInt(timerange.End(), 10))
+	params.Set("resolution", bluefloodResolution(timerange.Resolution()))
```
Nathan-Fenner: I'm not sure about these changes, exactly, since I'm not sure what units are involved in the blueflood backend portion. `.Start()` and `.End()` should be interpreted as milliseconds.

jeeyoungk: yes, they should be interpreted as milliseconds, so this is a good change.

jeeyoungk:
LGTM

api/types.go
achow: I can see that. Not sure about efficiency (or whether we need to worry about it), but whatever is idiomatic should be the right way.

PR #59, opened 2015-06-09
integration tests

jeeyoungk:


query/command_test.go
```diff
@@ -70,3 +104,71 @@ func TestCommand_Describe(t *testing.T) {
...
+		command := rawCommand.(*SelectCommand)
+		rawResult, err := command.Execute(fakeBackend)
+		if err != nil {
+			if !test.expectError {
+				a.Errorf("Unexpected error while executing: %s", err.Error())
+			}
+		} else {
+			casted := rawResult.([]value)
+			first, _ := casted[0].toSeriesList(api.Timerange{})
+			a.Eq(first.Series[0].Values, test.expected.Series[0].Values)
```
Nathan-Fenner: Should probably have a `.FloatEq()` which allows a small amount of error (like `1e-10`), or it might be hard to test things involving division if they give values which have some numerical error.

PR #60, opened 2015-06-09
Fill in timeseries with NaN for missing data (Blueflood backend)

Nathan-Fenner:
Fills in missing data with `NaN`, so that all response series have `len(Values) == timerange.Slots()`. Also combines things with buckets if there's more than one datapoint returned for a given interval.

@jeeyoungk @achow

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -200,12 +200,17 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 		return nil, err
 	}
 
-	// Construct a Timeseries from the result
+	// Construct a Timeseries from the result:
 
+	// buckets are each filled with from the points stored in result.Values, according to their timestamps.
 	buckets := bucketsFromMetricPoints(result.Values, selectResultField, timerange)
 
+	// values will hold the final values to be returned as the series.
 	values := make([]float64, timerange.Slots())
 
+	// bucketSamplers is a map describing how to transform { bucket []float64 => values[i] float64 }.
+	// each function may assume that the bucket it is given is non-empty (so indexing 0 is valid),
+	// which is enforced by the caller (who substitutes NaN for such empty buckets).
 	bucketSamplers := map[api.SampleMethod]func([]float64) float64{
 		api.SampleMean: func(bucket []float64) float64 {
 			value := 0.0
@@ -214,7 +219,7 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
 			}
 			return value / float64(len(bucket))
 		},
-		api.SampleMin: func(bucket []float64) float64 { // These functions may assume bucket is non-empty.
+		api.SampleMin: func(bucket []float64) float64 {
 			value := bucket[0]
 			for _, v := range bucket {
 				value = math.Min(value, v)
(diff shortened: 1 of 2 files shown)


api/backend/blueflood/blueflood.go
```diff
@@ -168,16 +169,32 @@ func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod ap
...
+	// The array is initialized to NaN.
+	for i := range values {
+		values[i] = math.NaN()
+	}
+
+	for _, metricPoint := range result.Values {
+		// TODO: check the result of .FieldByName against nil (or a panic will occur)
+		value := reflect.ValueOf(metricPoint).FieldByName(selectResultField).Float()
+		// TODO: check the result of .FieldByName against nil (or a panic will occur)
+		timestamp := reflect.ValueOf(metricPoint).FieldByName("Timestamp").Int()
```
jeeyoungk: should just do `.Timestamp` - the only reason why we're doing reflection above is because the above can retrieve one of the many values.

jeeyoungk:
let's pull this out to a separate function, and write a simple test for it.

api/backend/blueflood/blueflood.go
Nathan-Fenner: It's still a JSON value object, so Go doesn't know about the `.Timestamp` field.

api/backend/blueflood/blueflood.go
```diff
@@ -115,17 +115,71 @@ func (b *Blueflood) FetchSeries(metric api.TaggedMetric, predicate api.Predicate
 	}, nil
 }
 
+func addMetricPoint(metricPoint MetricPoint, field func(MetricPoint) float64, timerange api.Timerange, buckets [][]float64) error {
+	value := field(metricPoint)
+	timestamp := metricPoint.Timestamp
+	// The index to assign within the array is computed using the timestamp.
+	// It rounds to the nearest index.
+	index := (timestamp - timerange.Start() + timerange.Resolution()/2) / timerange.Resolution()
```
jeeyoungk: let's just floor it.

api/backend/blueflood/blueflood.go
```diff
@@ -115,17 +115,71 @@ func (b *Blueflood) FetchSeries(metric api.TaggedMetric, predicate api.Predicate
 	}, nil
 }
 
+func addMetricPoint(metricPoint MetricPoint, field func(MetricPoint) float64, timerange api.Timerange, buckets [][]float64) error {
```
jeeyoungk: i might even return `bool`

api/backend/blueflood/blueflood.go
```diff
@@ -115,17 +115,71 @@ func (b *Blueflood) FetchSeries(metric api.TaggedMetric, predicate api.Predicate
...
+}
+
+func bucketsFromMetricPoints(metricPoints []MetricPoint, resultField func(MetricPoint) float64, timerange api.Timerange) [][]float64 {
+	buckets := make([][]float64, timerange.Slots())
+	// Make the buckets:
+	for i := range buckets {
+		buckets[i] = []float64{}
+	}
+	for _, point := range metricPoints {
+		addMetricPoint(point, resultField, timerange, buckets)
```
jeeyoungk: technically we can log it on failure? (but let's not do it now).

api/backend/blueflood/blueflood.go
```diff
@@ -115,17 +115,71 @@ func (b *Blueflood) FetchSeries(metric api.TaggedMetric, predicate api.Predicate
...
+	}
+	return buckets
+}
+
 func (b *Blueflood) fetchSingleSeries(metric api.GraphiteMetric, sampleMethod api.SampleMethod, timerange api.Timerange) ([]float64, error) {
 	// Use this lowercase of this as the select query param. Use the actual value
 	// to reflect into result MetricPoints to fetch the correct field.
-	var selectResultField string
+	var (
```
jeeyoungk: store this in `map[sampleMethod]HelperStruct`, where `HelperStruct` has the previous 3 fields?

PR #61, opened 2015-06-09
making integration tests of commands better by being more lenient on floating point errors

jeeyoungk:
@Nathan-Fenner

### DIFF
--- a/assert/assert.go
+++ b/assert/assert.go
@@ -90,6 +90,14 @@ func (assert Assert) EqFloat(actual, expected float64) {
 	}
 }
 
+// EqFloat fails the test if two floats aren't equal. NaNs are considered equal.
+func (assert Assert) EqApproximate(actual, expected, epsilon float64) {
+	delta := actual - expected
+	if !(-epsilon < delta && delta < epsilon) {
+		assert.withCaller("Expected=[%f], actual=[%f]", expected, actual)
+	}
+}
+
 // Eq fails the test if two arguments are not equal.
 func (assert Assert) Eq(actual, expected interface{}) {
 	if !reflect.DeepEqual(actual, expected) {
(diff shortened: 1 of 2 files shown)


assert/assert.go
```diff
@@ -90,6 +90,14 @@ func (assert Assert) EqFloat(actual, expected float64) {
 	}
 }
 
+// EqFloat fails the test if two floats aren't equal. NaNs are considered equal.
+func (assert Assert) EqApproximate(actual, expected, epsilon float64) {
+	delta := actual - expected
+	if !(-epsilon < delta && delta < epsilon) {
```
Nathan-Fenner: If `expected` is `NaN` or infinite  then this will say that they differ (always). A special case for this at the top of the function would be useful.

jeeyoungk: should feel like they should always differ, since ... what does it mean to be epsilon away from these values anyway?

assert/assert.go
```diff
@@ -90,6 +90,14 @@ func (assert Assert) EqFloat(actual, expected float64) {
 	}
 }
 
+// EqFloat fails the test if two floats aren't equal. NaNs are considered equal.
+func (assert Assert) EqApproximate(actual, expected, epsilon float64) {
+	delta := actual - expected
```
achow: Can probably `math.Abs()` this.

Why don't we just add this to `EqFloat()` with epsilon 1e-10? Don't think we really lose anything, and this ends up being what it's used for anyways (comparing floats).

assert/assert.go
Nathan-Fenner: It would mean that you couldn't use `.ExApproximate` if you've got a testcase whose results could possibly be `NaN` or `Inf` (and are expected) since they'd fail no matter what was returned, and you'd have to split into another test case. This wouldn't be a problem with single-value returns, but when returning a list whose results are approximate (but might contain `NaN`) then you wouldn't be able to use this function (alone).

Nathan-Fenner:
LGTM

PR #65, opened 2015-06-10
Aggregators handle missing data (by ignoring NaNs)

Nathan-Fenner:
Missing data will be ignored during aggregation. For example, if asked to aggregate 5 timeseries (at some particular time `t`), and the values are `{1, 3, 0, NaN, 6}`, the resulting value will be `10.0/4 = 2.5`. If all values are missing `{NaN, NaN, NaN, NaN, NaN}` then the aggregate result will also be `NaN`.

@jeeyoungk @achow


jeeyoungk:
test for it?

PR #72, opened 2015-06-11
Updating loggers so that it can be controlled & pluggable

jeeyoungk:


achow:
LGTM. Eventually we should use the glog interface or something similar. Info+ should default log in a normal environment.

PR #77, opened 2015-06-11
Remember query in url #hash

Nathan-Fenner:
When the user presses the submit button, their query is saved in the hash of the URL, such as:

`http://localhost:8080/static/#select%20metric%20from%20'2015%2F6%2F10%2015%20PST'%20to%20'2015%2F6%2F10%2016%20PST'`

The URL can be copied/pasted and followed by other users. When opening a page with an encoded query, it is run immediately so that the results can be viewed on the page without having to press "submit".

Other additions:
* Error messages are shown
* While waiting for data to come back, shows a message
* While waiting for the chart to render, shows a message
* Cancels old request when a new one is made (if the request hadn't finished yet).

@jeeyoungk @achow

### DIFF
--- a/main/static/script.js
+++ b/main/static/script.js
@@ -8,7 +8,7 @@ module.controller("mainCtrl", function($scope, $http) {
 	$scope.queryResult = "";
 	$scope.queryText = "";
 	$scope.onSubmitQuery = function() {
-		window.location.hash = "#" + $scope.queryText;
+		window.location.hash = "#" + encodeURIComponent($scope.queryText);
 		$http.get('/query', {params:{query: $scope.queryText}}).
 		  success(function(data, status, headers, config) {
 		    $scope.queryResult = data;
--- a/main/static/script.js
+++ b/main/static/script.js
@@ -23,14 +23,14 @@ module.controller("mainCtrl", function($scope, $http, $q) {
 			return;
 		}
 		var urlQuery = window.location.hash
-		if (urlQuery != "") {
+		if (urlQuery != "" && urlQuery != "#") {
 			// Drop the leading '#'
 			urlQuery = urlQuery.substring(1);
 			// The remainder is the query to perform.
 			// Decode it (if neccesarry).
 			$scope.queryText = decodeURIComponent(urlQuery);
+			launchRequest($scope.queryText);
 		}
-		launchRequest($scope.queryText);
 	}
 
 	function cancelQuery() {
--- a/main/static/script.js
+++ b/main/static/script.js
@@ -6,22 +6,16 @@ var chartsReady = false;
 
 var debounce = false;
 
-module.controller("mainCtrl", function($scope, $http) {
+module.controller("mainCtrl", function($scope, $http, $q) {
 	$scope.queryResult = "";
 	$scope.queryText = "";
+	$scope.launchedQuery = null;
+	// Triggers when the button is clicked.
 	$scope.onSubmitQuery = function() {
 		debounce = true;
 		window.location.hash = "#" + encodeURIComponent($scope.queryText);
-		$http.get('/query', {params:{query: $scope.queryText}}).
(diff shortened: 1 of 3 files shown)


main/static/script.js
```diff
@@ -1,13 +1,16 @@
...
 
+var chartsReady = false;
+
+
+
 module.controller("mainCtrl", function($scope, $http) {
 	$scope.queryResult = "";
 	$scope.queryText = "";
 	$scope.onSubmitQuery = function() {
+		window.location.hash = "#" + encodeURIComponent($scope.queryText);
```
jeeyoungk: oh, you should use `$location` and `$path` - those provide a bit better abstraction over the current locations.

main/static/script.js
```diff
@@ -16,20 +19,47 @@ module.controller("mainCtrl", function($scope, $http) {
...
+			// Decode it (if neccesarry).
+			$scope.queryText = decodeURIComponent(urlQuery);
+			$scope.onSubmitQuery();
+		}
+	}
+
+	readHash(); // Read the current hash (if present).
+
+	// Whenever the user changes the hash (such as pasting in a new URL), call readHash() again.
+	window.addEventListener("hashchange", readHash, false);
```
jeeyoungk: we shouldn't need to listen for the URL change, because the page will refresh.

main/static/script.js
```diff
@@ -16,20 +19,47 @@ module.controller("mainCtrl", function($scope, $http) {
...
 function dateFromIndex(index, timerange) {
 	return new Date(timerange.start + timerange.resolution * index);
 }
 
+
+
 function resultUpdate(object) {
+	if (!chartsReady) {
+		// Ask again in 30ms.
+		setTimeout(function(){resultUpdate(object);}, 10);
```
jeeyoungk: we should probably queue this up in some array, and call this on `setOnLoadCallback`.

main/static/script.js
Nathan-Fenner: With anchors, it actually doesn't (which is why this is necesarry, but wouldn't be needed if I used `?q=` parameters). For example, going to http://en.wikipedia.org/wiki/Water and clicking on the table of contents doesn't cause the page to reload, even though the URL changed (since only the hash part changed).

Nathan-Fenner: Also, the same holds true even if the user types it in themselves, at least if the hash is already present in the URL.

jeeyoungk: ahh, so you're storing in the hash-part of the URL. sorry i got a bit confused.

main/static/script.js
```diff
@@ -1,13 +1,17 @@
...
+var chartsReady = false;
+
+var debounce = false;
+
 module.controller("mainCtrl", function($scope, $http) {
 	$scope.queryResult = "";
 	$scope.queryText = "";
 	$scope.onSubmitQuery = function() {
+		debounce = true;
+		window.location.hash = "#" + encodeURIComponent($scope.queryText);
```
jeeyoungk: can we use `$location` instead?

main/static/script.js
```diff
@@ -1,13 +1,17 @@
 "use strict";
 
-console.log("Hello, world!");
-
 var module = angular.module("main",[]);
 
+var chartsReady = false;
+
+var debounce = false;
```
jeeyoungk: what's the intent you're trying to capture here? (I think I understand the code but not too sure).

Nathan-Fenner: When `onSubmitQuery` is called, it changes the hash. So the `hashchange` event fires, which calls `onSubmitQuery` again; both calls to `onSubmitQuery` fetch from the database.

So the `debounce` tells `hashchange` event listener not to call `onSubmitQuery` so that this second, extraneous call doesn't occur.

jeeyoungk:
LGTM

PR #79, opened 2015-06-11
Parallel fetch

Nathan-Fenner:
For multiple-series fetches, performs the fetch in parallel, with a limited number of active fetches at a time. (Currently, 4).

@jeeyoungk @achow

### DIFF
--- a/query/expression.go
+++ b/query/expression.go
@@ -26,20 +26,19 @@ import (
 const simultaneousFetchLimit = 4
 
 // fetchTicketsQueue holds tickets, which a worker must pick up in order to make a query.
-var fetchTicketsQueue = make(chan bool, simultaneousFetchLimit)
+var fetchTicketsQueue = make(chan struct{}, simultaneousFetchLimit)
 
 // init() adds all the tickets needed to the fetchTicketsQueue.
 func init() {
 	for i := 0; i < simultaneousFetchLimit; i++ {
-		fetchTicketsQueue <- true
+		fetchTicketsQueue <- struct{}{}
 	}
 }
 
 // fetchLazy issues a goroutine to compute the timeseries once a fetchticket becomes available.
 // It returns a channel to wait for the response to finish (the error).
 // It stores the result of the function invokation in the series pointer it is given.
-func fetchLazy(result *api.Timeseries, fun func() (api.Timeseries, error)) chan error {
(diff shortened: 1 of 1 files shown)


query/expression.go
```diff
@@ -23,6 +23,57 @@ import (
...
+		channel <- err
+	}()
+	return channel
+}
+
+// fetchManyLazy abstracts upon fetchLazy so that looping over the resulting channels is not needed.
+// It returns any overall error, as well as a slice of the resulting timeseries.
+func fetchManyLazy(funs []func() (api.Timeseries, error)) ([]api.Timeseries, error) {
+	results := make([]api.Timeseries, len(funs))
+	channels := make([]chan error, len(funs))
```
jeeyoungk: errors:= `make(chan error, len(funs))`

query/expression.go
```diff
@@ -23,6 +23,55 @@ import (
 	"github.com/square/metrics/query/aggregate"
 )
 
+const simultaneousFetchLimit = 4
```
Nathan-Fenner: I'm not sure if it's safe to move this as a parameter into the evaluation context, since it's possible for many expressions to be evaluating simultaneously. (It's still pretty safe, just slightly less so, taking 4 times fewer concurrent expressions than the solution without threading).

jeeyoungk: it should be some option that can be controllable per app. we can pair on it later to extract it out correctly.

jeeyoungk: the one i dislike about `simultaneousFetchLimit` and more importantly `fetchTicketsQueue` being in file scope is that it's a global variable shared amongst everything. (as unlikely as it is) we can have multiple `api`s and `backend`s initialized and yet we are limiting the concurrency.

there are ways to limit this to be api-scoped or backend-scoped (i.e. enforce the limit per backend / api). it would be stored in some state (let's call this type `Environment`) for now, and would be shared amongst all the query execution.

Currently we don't have such an object - command has a member function with the signature `Execute(api.Backend, api.API)` and both Backend and API doesn't allow you to store such a state, but we can imagine passing a third parameter 

on the other hand, I realized that ticketing could've been implemented as a wrapper backend. a wrapper that takes an actual backend as an input, and just implements the ticker system in front of it. then we'll implement per-backend concurrency control without having a global variable.

PR #80, opened 2015-06-11
Add relative timestamp support

achow:
@jeeyoungk @Nathan-Fenner

### DIFF
--- a/query/parser.go
+++ b/query/parser.go
@@ -105,7 +105,7 @@ var relativeTimeRegexp = regexp.MustCompile(`^-(\d+)([a-zA-Z])$`)
 // parseRelativeTime transforms a relative time string (e.g. `-10m`) to a
 // timestamp in millis.
 func parseRelativeTime(s string, now time.Time) (int64, error) {
-	if s == "now" {
+	if strings.ToLower(s) == "now" {
 		return now.Unix() * 1000, nil
 	}
 
@@ -389,7 +389,6 @@ func (p *Parser) insertPropertyKeyValue() {
 		if unix, err = parseDate(value); err == nil {
 			// Valid, so do nothing.
 		} else if unix, err = parseRelativeTime(value, time.Now().UTC()); err != nil {
-			fmt.Printf("`%s` is invalid: %s\n", value, err.Error())
 			p.flagSyntaxError(SyntaxError{
 				token:   value,
 				message: err.Error(),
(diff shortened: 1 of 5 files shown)


query/parser.go
```diff
@@ -99,6 +99,50 @@ var dateFormats = []string{
 	time.RFC822Z,
 }
 
+// `now` isn't captured in this regex
+var relativeTimeRegexp = regexp.MustCompile(`^-(\d+)([a-zA-Z])$`)
+
+// parseRelativeTime transforms a relative time string (e.g. `-10m`) to a
+// timestamp in millis.
+func parseRelativeTime(s string, now time.Time) (int64, error) {
+	if s == "now" {
```
jeeyoungk: s can be mixed case :)

Nathan-Fenner:
LGTM

jeeyoungk:
LGTM

query/parser.go
```diff
@@ -340,8 +384,12 @@ func (p *Parser) insertPropertyKeyValue() {
 	case "from":
 		fallthrough
 	case "to":
-		unix, err := parseDate(value)
-		if err != nil {
+		var unix int64
+		var err error
+		if unix, err = parseDate(value); err == nil {
+			// Valid, so do nothing.
+		} else if unix, err = parseRelativeTime(value, time.Now().UTC()); err != nil {
```
jeeyoungk: Actually, I see this being somewhat problematic, (even though it mostly won't)

1. `now` will be slightly different between `from` and `to`, because we're taking unix timestamps. these values should be minimal but it's still a bit annoying.
2. if there's a delay between parsing & execution, then "now" will stay the same. in fact, it is a valid question whether if we would want be able to execute this query over and over again with the same or different timestamp for `now`.

Nathan-Fenner: To address (1) you could compute `time.Now()` just once, prior to construction of the `EvaluationContext`, and then pass it as a parameter here.

For (2) you could do something similar, only allow an override for `now` specified by the client/query (who calculate it the first time, then save the query alongside the "now" timestamp to reproduce the original query).

achow: In practice, probably not a big deal. There might be a difference of 1 second between what `from` and `to` think `now` is.

For replayability, just log the timestamps when queries run or don't use the feature. It's syntactic sugar for running commands on the fly.

jeeyoungk: Yes.

I think what we currently have is good.

PR #82, opened 2015-06-12
String literal expression

Nathan-Fenner:
Adds string literal nodes to the parse tree.

Also adds tests for `timeshift(...)`.
@jeeyoungk @achow


jeeyoungk:
LGTM

PR #83, opened 2015-06-12
Move multibackend instantiation into main/ui.go out of ui/ui.go

Nathan-Fenner:
This change means that only one `MultiBackend` is created for the duration of a life of the program, rather than one per command.

@jeeyoungk @achow

### DIFF
--- a/main/ui.go
+++ b/main/ui.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 
+	"github.com/square/metrics/api/backend"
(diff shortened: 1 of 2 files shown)


main/ui.go
```diff
@@ -27,6 +28,6 @@ func main() {
 	common.SetupLogger()
 
 	apiInstance := common.NewAPI()
```
jeeyoungk: why not parallel? :)

jeeyoungk:
LGTM

PR #84, opened 2015-06-12
Limit the number of fetches

Nathan-Fenner:
An `ExecutionContext` type has been created which replaces the `(API, MultiBackend)` passed to each `.Execute(...)` call.

Before a fetch is performed, the number available is checked against the number of series desired. If not enough are available, the metric fetching expression returns an error.

@jeeyoungk @achow

### DIFF
--- a/query/expression.go
+++ b/query/expression.go
@@ -34,7 +34,8 @@ type EvaluationContext struct {
 	API          api.API          // Api to obtain metadata from
 	Timerange    api.Timerange    // Timerange to fetch data from
 	SampleMethod api.SampleMethod // SampleMethod to use when up/downsampling to match the requested resolution
-	Predicate    api.Predicate
+	Predicate    api.Predicate    // Predicate to apply to TagSets prior to fetching
+	FetchLimit   *int             // A limit on the number of fetches which may be performed
 }
 
 // Expression is a piece of code, which can be evaluated in a given
@@ -79,6 +80,12 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
 		return nil, err
 	}
 	filtered := applyPredicates(metricTagSets, predicate)
+
+	*context.FetchLimit -= len(filtered)
+
+	if *context.FetchLimit < 0 {
+		return nil, errors.New("fetch limit exceeded: too many series to fetch")
+	}
 
 	metrics := make([]api.TaggedMetric, len(filtered))
 	for i := range metrics {
--- a/query/expression.go
+++ b/query/expression.go
@@ -35,7 +35,30 @@ type EvaluationContext struct {
 	Timerange    api.Timerange    // Timerange to fetch data from
 	SampleMethod api.SampleMethod // SampleMethod to use when up/downsampling to match the requested resolution
 	Predicate    api.Predicate    // Predicate to apply to TagSets prior to fetching
-	FetchLimit   *int             // A limit on the number of fetches which may be performed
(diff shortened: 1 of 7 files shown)


query/expression.go
```diff
@@ -80,6 +81,12 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
 	}
 	filtered := applyPredicates(metricTagSets, predicate)
 
+	*context.FetchLimit -= len(filtered)
+
+	if *context.FetchLimit < 0 {
```
jeeyoungk: let's have a test around this case.

query/expression.go
```diff
@@ -34,7 +34,8 @@ type EvaluationContext struct {
 	API          api.API          // Api to obtain metadata from
 	Timerange    api.Timerange    // Timerange to fetch data from
 	SampleMethod api.SampleMethod // SampleMethod to use when up/downsampling to match the requested resolution
-	Predicate    api.Predicate
+	Predicate    api.Predicate    // Predicate to apply to TagSets prior to fetching
+	FetchLimit   *int             // A limit on the number of fetches which may be performed
```
achow: Why is this `*int` instead of just `int`?

achow: Nvm, I kind of see what's going on here..

Nathan-Fenner: By leaving it as an `int*`, only the code that actually cares about it (namely, metric fetching) needs to modify it. If it were an `int`, then every single recursive expression `.Evaluate()` call would need to return an extra parameter (either the entire context, or just the number of things consumed).

achow: I think it might be more idiomatic to ship around `*EvaluationContext`. Feels weird to be propagating updates with a pointer in a something passed by value... @jeeyoungk ?

jeeyoungk: yes, we (currently) want to share a single value amongst the entire execution, but we're copying the struct around thus we need to have an integer pointer.

we may need to make it into an atomic integer to avoid race conditions. Right now, we don't implement it, but i can see implementing parallel fetches amongst subexpressions. i.e. if you have 

```
select cpu_user + cpu_kernel where ...
```
you can fetch the left-hand-side and right-hand-side concurrently. Right now, our parallel fetching logic only fetches in-parallel within a single fetch expression.

jeeyoungk: @achow oh, i didn't read the last comment.

i think in sort term, this being pointer is a better solution because there are few stateful and stateless values in this object. (timerange for timeshift for example). being able to safely copy and make modifications to it is a good property to have.

if we make `EvaluationContext` as a pointer (and keep the limit as value), then we have to re-evaluate how timeshifting needs to be done - because we cannot simply modify it, since it's a global object.

in future, i feel like it's worthwhile to categorize fields that can freely be changed per sub-expression node and data that must be shared between expressions.

query/expression.go
```diff
@@ -34,7 +34,31 @@ type EvaluationContext struct {
...
+	return f
+}
+
+func (c fetchCounter) Consume(n int) bool {
+	<-c.ticket
+	*c.count -= n
+	r := *c.count < 0
+	c.ticket <- struct{}{}
+	return r
 }
```
Nathan-Fenner: This function performs the decrement / compare in a thread-safe manner (it can't be interrupted because the channel will block for all other threads). So now it's safe to make the rest of evaluation concurrent without causing any problems here.

jeeyoungk:
LGTM

PR #86, opened 2015-06-13
Better structured backend error codes.

jeeyoungk:
This will allow us (in future) to act against different errors.

* implement blueflood to use these error codes.
* refactor blueflood to not expose these fields.

@alexc @Nathan-Fenner


Nathan-Fenner:
LGTM

PR #91, opened 2015-06-15
Unify durations 

Nathan-Fenner:
When using relative dates with `parseDate` (which now asks for a `now` time), the format is identical to `parseDuration`:

The units supported are `s`, `m`, `h`, `hr`, `d`, `w` (week = 7 days), `M`, `mo`, (month = 30 days), `y`, `yr`, (year = 365 days).

### DIFF
--- a/main/query.go
+++ b/main/query.go
@@ -37,7 +37,7 @@ func main() {
 	}
 
 	apiInstance := common.NewAPI()
-	myBackend := blueflood.NewBlueflood(*common.BluefloodUrl, *common.BluefloodTenantId)
+	myBackend := blueflood.NewBlueflood(blueflood.BluefloodClientConfig{*common.BluefloodUrl, *common.BluefloodTenantId})
 
 	l := liner.NewLiner()
 	defer l.Close()
--- a/query/parser.go
+++ b/query/parser.go
@@ -402,13 +402,19 @@ func (p *Parser) insertPropertyKeyValue() {
 	case "resolution":
 		// The value must be determined to be an int if the key is "resolution".
 		intValue, err := strconv.ParseInt(value, 10, 64)
-		if err != nil {
-			p.flagSyntaxError(SyntaxError{
-				token:   value,
-				message: fmt.Sprintf("Expected number but parse failed; %s", err.Error()),
-			})
+		if err == nil {
+			contextNode.Resolution = intValue
+			break
 		}
-		contextNode.Resolution = intValue
+		duration, err := toDuration(stringValue(value))
+		if err == nil {
+			contextNode.Resolution = duration
+			break
+		}
+		p.flagSyntaxError(SyntaxError{
+			token:   value,
+			message: fmt.Sprintf("Expected number but parse failed; %s", err.Error()),
+		})
 	default:
 		p.flagSyntaxError(SyntaxError{
 			token:   key,
--- a/query/parser.go
+++ b/query/parser.go
@@ -102,6 +102,11 @@ var dateFormats = []string{
 // parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
 func parseDate(date string, now time.Time) (int64, error) {
 
+	// Millisecond epoch timestamp.
+	if epoch, err := strconv.ParseInt(date); err != nil {
+		return epoch, nil
+	}
+
 	relativeTime, err := toDuration(stringValue(date))
 	if err != nil {
 		// A relative date.
@@ -359,9 +364,7 @@ func (p *Parser) insertPropertyKeyValue() {
 				message: fmt.Sprintf("Expected sampling method 'max', 'min', or 'mean' but got %s", value),
 			})
 		}
-	case "from":
-		fallthrough
-	case "to":
+	case "from", "to":
 		var unix int64
 		var err error
 		now := time.Now()
--- a/query/parser.go
+++ b/query/parser.go
@@ -99,7 +99,10 @@ var dateFormats = []string{
 	time.RFC822Z,
 }
 
-var relativeTimeRegexp = regexp.MustCompile(`^now\s*([+-])\s*([0-9]+)([smhdwMy])$`)
+var (
+	relativeTimeRegexp      = regexp.MustCompile(`^now\s*([+-])\s*([0-9]+)([smhdwMy])$`)
+	relativeTimeRegexpShort = regexp.MustCompile(`(-)\s*([0-9]+)([smhdwMy])$`)
+)
 
 // parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
 func parseDate(date string) (int64, error) {
@@ -109,6 +112,10 @@ func parseDate(date string) (int64, error) {
 	}
 
 	matches := relativeTimeRegexp.FindStringSubmatch(date)
+	if matches == nil {
+		// If the long match fails, try to short match.
+		matches = relativeTimeRegexpShort.FindStringSubmatch(date)
+	}
 	if matches != nil {
 		// This match can be used to determine the duration of time.
 		now := time.Now().Unix()
--- a/query/parser.go
+++ b/query/parser.go
@@ -99,7 +99,7 @@ var dateFormats = []string{
 	time.RFC822Z,
 }
 
-var relativeTimeRegexp = regexp.MustCompile(`^now\s*([+-])\s*(\d+)([smhdMy])$`)
+var relativeTimeRegexp = regexp.MustCompile(`^now\s*([+-])\s*([0-9]+)([smhdwMy])$`)
 
 // parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
 func parseDate(date string) (int64, error) {
@@ -109,7 +109,7 @@ func parseDate(date string) (int64, error) {
 	}
 
 	matches := relativeTimeRegexp.FindStringSubmatch(date)
-	if matches == nil {
+	if matches != nil {
 		// This match can be used to determine the duration of time.
 		now := time.Now().Unix()
 		sign := int64(1)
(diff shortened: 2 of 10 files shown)


query/parser.go
```diff
@@ -99,57 +99,55 @@ var dateFormats = []string{
...
 		}
-
-		offset := offsetSeconds * 1000
-		specifier := matches[2]
-
-		switch specifier {
+		switch matches[3] {
 		case "s":
-			// Already in millis, do nothing.
+			offset *= 1
```
achow: This does nothing. Should just leave it as-is.

query/parser.go
```diff
@@ -99,57 +99,55 @@ var dateFormats = []string{
...
+		case "w":
+			offset *= 60 * 60 * 24 * 7
 		case "M":
 			offset *= 60 * 60 * 24 * 30
 		case "y":
 			offset *= 60 * 60 * 24 * 365
 		default:
-			return -1, fmt.Errorf("invalid relative timestamp %s", s)
+			// Won't happen, regex can't produce any other value.
+			panic(matches[3])
```
achow: error, don't panic. This is recoverable, especially if we're using a REPL.

Nathan-Fenner: The code is unreachable, provided the timestamp regexp (and the switch) is correct/consistent, so I decided to panic.

achow:
Why require 'now - XX'? This is meant to be syntactic sugar, otherwise it just adds more friction when you're pulling metrics from the UI or otherwise building queries.

Nathan-Fenner:
TODO: allow `'-10h'` as a valid relative timestamp

(originally left out due to (erroneous) concerns regarding conflicts between `parseDuration` and `parseDate`, now resolved)

Nathan-Fenner:
Now short relative timestamps (`'-10h'`) are supported again.

query/parser.go
```diff
@@ -99,57 +99,62 @@ var dateFormats = []string{
...
+	relativeTimeRegexpShort = regexp.MustCompile(`(-)\s*([0-9]+)([smhdwMy])$`)
+)
 
-// parseRelativeTime transforms a relative time string (e.g. `-10m`) to a
-// timestamp in millis.
-func parseRelativeTime(s string, now time.Time) (int64, error) {
-	if strings.ToLower(s) == "now" {
-		return now.Unix() * 1000, nil
+// parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
+func parseDate(date string) (int64, error) {
```
achow: Why remove `now`? Do tests even pass?

achow:
If we support both `now - 10h` and `-10h`, what are the benefits of having the former?

Nathan-Fenner:
There isn't much advantage. I'll pull the longer way out, and unify with duration parsing a bit more.

query/parser.go
```diff
@@ -99,57 +99,60 @@ var dateFormats = []string{
...
 
-// `now` isn't captured in this regex
-var relativeTimeRegexp = regexp.MustCompile(`^-(\d+)([a-zA-Z])$`)
+var (
+	relativeTimeRegexp      = regexp.MustCompile(`^now\s*([+-])\s*([0-9]+)([smhdwMy])$`)
+	relativeTimeRegexpShort = regexp.MustCompile(`(-)\s*([0-9]+)([smhdwMy])$`)
+)
+
+// parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
+func parseDate(date string, now time.Time) (int64, error) {
```
achow: If you keep the relative time conversion as a separate function, you can reuse it for parsing the `resolution` duration.

query/value.go
```diff
@@ -88,6 +89,8 @@ func (value scalarValue) toScalar() (float64, error) {
 	return float64(value), nil
 }
 
+var durationRegexp = regexp.MustCompile(`^([+-]?[0-9]+)([smhdwMy])$`)
```
achow: Can probably do without the `+` since omission is implicitly positive.

query/parser.go
```diff
@@ -99,57 +99,16 @@ var dateFormats = []string{
...
-		case "y":
-			offset *= 60 * 60 * 24 * 365
-		default:
-			return -1, fmt.Errorf("invalid relative timestamp %s", s)
-		}
+// parseDate converts the given datestring (from one of the allowable formats) into a millisecond offset from the Unix epoch.
+func parseDate(date string, now time.Time) (int64, error) {
 
-		return now.Unix()*1000 - offset, nil
+	relativeTime, err := toDuration(stringValue(date))
```
achow: Should ensure this duration isn't positive, since queries into the future don't really make sense.

query/value.go
```diff
@@ -96,9 +99,30 @@ func toDuration(value value) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	duration, err := time.ParseDuration(timeString)
+	matches := durationRegexp.FindStringSubmatch(timeString)
+	if matches == nil {
+		return 0, fmt.Errorf("expected duration to be of the form `[0-9]+[smhdwMy]`")
+	}
+	duration, err := strconv.ParseInt(matches[1], 10, 0)
```
achow: Needs to handle negatives.

query/parser.go
```diff
@@ -402,9 +364,10 @@ func (p *Parser) insertPropertyKeyValue() {
 	case "to":
 		var unix int64
 		var err error
-		if unix, err = parseDate(value); err == nil {
+		now := time.Now()
+		if unix, err = parseDate(value, now); err == nil {
```
achow: This should turn into a single `err != nil` since there's no second `if` anymore.

query/value.go
Nathan-Fenner: Handle in what way?

query/parser.go
```diff
@@ -418,13 +381,19 @@ func (p *Parser) insertPropertyKeyValue() {
 	case "resolution":
 		// The value must be determined to be an int if the key is "resolution".
 		intValue, err := strconv.ParseInt(value, 10, 64)
-		if err != nil {
-			p.flagSyntaxError(SyntaxError{
-				token:   value,
-				message: fmt.Sprintf("Expected number but parse failed; %s", err.Error()),
-			})
+		if err == nil {
```
achow: See if you can turn this into a chain of `if`s instead of `break`s.

query/value.go
```diff
@@ -88,17 +89,42 @@ func (value scalarValue) toScalar() (float64, error) {
...
+	case "h":
+		scale = 1000 * 60 * 60
+	case "d":
+		scale = 1000 * 60 * 60 * 24
+	case "w":
+		scale = 1000 * 60 * 60 * 24 * 7
+	case "M":
+		scale = 1000 * 60 * 60 * 24 * 30
+	case "y":
+		scale = 1000 * 60860 * 24 * 365
```
achow: Where does 60860 come from?

Nathan-Fenner: Typo, and good catch. 8 is asterisk key.

main/query.go
```diff
@@ -37,7 +37,7 @@ func main() {
 	}
 
 	apiInstance := common.NewAPI()
-	myBackend := blueflood.NewBlueflood(*common.BluefloodUrl, *common.BluefloodTenantId)
+	myBackend := blueflood.NewBlueflood(blueflood.BluefloodClientConfig{*common.BluefloodUrl, *common.BluefloodTenantId})
```
jeeyoungk: this _shouldn't_ be needed if we rebase.

Nathan-Fenner: It looks like `main/query.go` got skipped over somehow, so it's not currently in master.

query/value.go
```diff
@@ -88,17 +89,42 @@ func (value scalarValue) toScalar() (float64, error) {
 	return float64(value), nil
 }
 
+var durationRegexp = regexp.MustCompile(`^([+-]?[0-9]+)([smhdwMy]|ms)$`)
```
jeeyoungk: this is slowly getting more complex :P

main/query.go
jeeyoungk: ahh :/

query/parser_test.go
```diff
@@ -35,16 +35,20 @@ func Test_parseRelativeTime(t *testing.T) {
 		{"-3m", 1413321686000, true},
 		{"-4h", 1413307466000, true},
 		{"-5d", 1412889866000, true},
+		{"-3w", 1411507466000, true},
```
achow: Needs a test for `ms`

query/value.go
```diff
@@ -88,17 +89,42 @@ func (value scalarValue) toScalar() (float64, error) {
...
+		scale = 1000 * 60 * 60
+	case "d":
+		scale = 1000 * 60 * 60 * 24
+	case "w":
+		scale = 1000 * 60 * 60 * 24 * 7
+	case "M":
+		scale = 1000 * 60 * 60 * 24 * 30
+	case "y":
+		scale = 1000 * 60 * 60 * 24 * 365
 	}
```
achow: Needs a default case.

query/value.go
```diff
@@ -88,17 +89,42 @@ func (value scalarValue) toScalar() (float64, error) {
...
 // It converts the return value to milliseconds.
 func toDuration(value value) (int64, error) {
 	timeString, err := value.toString()
 	if err != nil {
-		return 0, err
+		return -1, err
+	}
+	matches := durationRegexp.FindStringSubmatch(timeString)
+	if matches == nil {
+		return -1, fmt.Errorf("expected duration to be of the form `[0-9]+[smhdwMy]`")
```
achow: You can `regexp.String()` to pull the regex out of the `Regexp`

query/parser_test.go
Nathan-Fenner: There's now exhaustive tests for suffixes in `value_test.go`

jeeyoungk:
LGTM on my side - @achow ?

PR #92, opened 2015-06-15
Add support for configured ttl, standardize on configs

achow:
@jeeyoungk @Nathan-Fenner

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/square/metrics/api"
 	"github.com/square/metrics/log"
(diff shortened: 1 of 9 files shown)


api/backend/blueflood/blueflood.go
```diff
@@ -68,14 +103,21 @@ type BluefloodClientConfig struct {
...
+	b.config.Ttls = map[Resolution]int64{}
+	for k, v := range c.Ttls {
+		b.config.Ttls[k] = v
+	}
+
+	return &b
 }
 
-func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Timeseries, error) {
+func (b *Blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Timeseries, error) {
```
jeeyoungk: why did you re-export it?

api/backend/blueflood/blueflood.go
```diff
@@ -218,18 +260,21 @@ var samplerMap map[api.SampleMethod]struct {
 
 // Blueflood keys the resolution param to a java enum, so we have to convert
 // between them.
-func bluefloodResolution(r int64) resolution {
+func (c Config) bluefloodResolution(resolution int64, timestamp int64) string {
```
jeeyoungk: i feel like this logic of:
```
given a desired resolution, timerange (we pass in current time), and current time, give us a resolution
```
can be refactored a bit more. specifically, i feel like we're going to use it when we implement graphite backend (or other different backend).

achow: I don't think making something like this generic is going to buy us much. It's only like 20 lines of code and used only in once place.

jeeyoungk:
LGTM

PR #95, opened 2015-06-16
Function registration

Nathan-Fenner:
No unit tests yet. Existing tests for expressions are still standing, however.

@jeeyoungk @achow

### DIFF
--- a/query/aggregate/aggregate.go
+++ b/query/aggregate/aggregate.go
@@ -74,7 +74,7 @@ func groupBy(list api.SeriesList, tags []string) []group {
 	return result
 }
 
-var aggregateMap = map[string]func([]float64) float64{
+var AggregateMap = map[string]func([]float64) float64{
 	"aggregate.sum":
 	// The aggregatefor sum finds the sum of the given array.
 	func(array []float64) float64 {
@@ -127,16 +127,16 @@ var aggregateMap = map[string]func([]float64) float64{
 
 // GetAggregate gives the aggregate of the given name (and false if it doesn't exist).
 func GetAggregate(name string) (func([]float64) float64, bool) {
-	aggregate, ok := aggregateMap[name]
+	aggregate, ok := AggregateMap[name]
 	return aggregate, ok
 }
 
 // AddAggregate adds an aggregate of a given name to the aggregate map.
 func RegisterAggregate(name string, aggregate func([]float64) float64) error {
-	if _, ok := aggregateMap[name]; ok {
+	if _, ok := AggregateMap[name]; ok {
 		return errors.New(fmt.Sprintf("aggregate %s has already been declared", name))
 	}
-	aggregateMap[name] = aggregate
+	AggregateMap[name] = aggregate
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


query/expression.go
```diff
@@ -23,6 +23,28 @@ import (
...
+	MustRegister(MakeAggregateMetricFunction("aggregate.sum", aggregate.AggregateMap["aggregate.sum"]))
+	// Transformations
+	MustRegister(MakeTransformMetricFunction("transform.derivative", 0, transformTable["transform.derivative"]))
+	MustRegister(MakeTransformMetricFunction("transform.integral", 0, transformTable["transform.integral"]))
+
+	MustRegister(MakeTransformMetricFunction("transform.rate", 0, transformTable["transform.rate"]))
+	MustRegister(MakeTransformMetricFunction("transform.cumulative", 0, transformTable["transform.cumulative"]))
+	MustRegister(MakeTransformMetricFunction("transform.moving_average", 0, transformTable["transform.moving_average"]))
+	MustRegister(MakeTransformMetricFunction("transform.default", 0, transformTable["transform.default"]))
+	MustRegister(MakeTransformMetricFunction("transform.abs", 0, transformTable["transform.abs"]))
```
Nathan-Fenner: I don't know if these should be here, or they should be in their corresponding files (`aggregate.go / `transform.go`).

jeeyoungk: i think them being here is nice (because we can see them). third-party functions should be registered from elsewhere.

query/aggregate/aggregate.go
```diff
@@ -74,7 +74,7 @@ func groupBy(list api.SeriesList, tags []string) []group {
 	return result
 }
 
-var aggregateMap = map[string]func([]float64) float64{
+var AggregateMap = map[string]func([]float64) float64{
```
jeeyoungk: i'm wondering whether this is even needed - because we don't need a map to store it right? Technically we can just have:

```
func Sum([]float64) float64 {}
...
func Mean([]float64) float64 {}
...
right?
```

query/expression.go
```diff
@@ -121,143 +144,17 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
...
-		if series, ok := value.(seriesListValue); ok {
-			// If it's a series, then we need to reset its timerange to the original.
-			// Although it's questionably useful to use timeshifting for a non-series,
-			// it seems sensible to allow it anyway.
-			series.Timerange = context.Timerange
-		}
-		return value, nil
+	fun, ok := GetFunction(expr.functionName)
+	if !ok {
+		return nil, fmt.Errorf("no such function %s", expr.functionName)
```
jeeyoungk: should have a error (maybe `SyntaxError`) associated with this.

query/function.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if seriesValue, ok := result.(seriesListValue); ok {
+			seriesValue.Timerange = context.Timerange
+			return seriesValue, nil
+		}
+		return result, nil
+
```
jeeyoungk: nit: whitespace.

PR #98, opened 2015-06-17
Implement Filter Functions

Nathan-Fenner:
This PR is based on #95 and adds on the definition of several "filtering" functions, which limit the number of returned Timeseries.

#95 should be finalized before this branch.

### DIFF
--- a/query/expression.go
+++ b/query/expression.go
@@ -55,7 +55,7 @@ func init() {
 				sum += x
 			}
 		}
-		return sum / count
+		return sum / float64(count)
 	}
 	MustRegister(MakeFilterMetricFunction("filter.highest_mean", meanSummary, false))
 	MustRegister(MakeFilterMetricFunction("filter.lowest_mean", meanSummary, true))
--- a/query/expression.go
+++ b/query/expression.go
@@ -46,7 +46,6 @@ func init() {
 	// Timeshift
 	MustRegister(TimeshiftFunction)
 	// Filter
-
 	MustRegister(MakeFilterMetricFunction("filter.highest_mean", aggregate.AggregateMean, false))
 	MustRegister(MakeFilterMetricFunction("filter.lowest_mean", aggregate.AggregateMean, true))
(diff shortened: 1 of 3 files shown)


query/expression.go
```diff
@@ -17,12 +17,72 @@ package query
...
+	MustRegister(MakeTransformMetricFunction("transform.integral", 0, transformTable["transform.integral"]))
+	MustRegister(MakeTransformMetricFunction("transform.rate", 0, transformTable["transform.rate"]))
+	MustRegister(MakeTransformMetricFunction("transform.cumulative", 0, transformTable["transform.cumulative"]))
+	MustRegister(MakeTransformMetricFunction("transform.moving_average", 0, transformTable["transform.moving_average"]))
+	MustRegister(MakeTransformMetricFunction("transform.default", 0, transformTable["transform.default"]))
+	MustRegister(MakeTransformMetricFunction("transform.abs", 0, transformTable["transform.abs"]))
+	// Timeshift
+	MustRegister(TimeshiftFunction)
+	// Filter
+	meanSummary := func(xs []float64) float64 {
```
jeeyoungk: i feel like these functions should be defined elsewhere, and `init()` only performs  registrations.

query/filter.go
```diff
@@ -0,0 +1,71 @@
...
+
+package query
+
+import (
+	"sort"
+
+	"github.com/square/metrics/api"
+)
+
+type filterList struct {
```
jeeyoungk: i actually really like this abstraction.

query/function.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("function %s in failed to register", fun.Name))
+	}
+}
+
+type MetricFunction struct {
+	Name        string
+	MinArgument int
+	MaxArgument int
+	Groups      bool // Whether the function allows a 'group by' clause.
```
jeeyoungk: nice

PR #99, opened 2015-06-17
Functions compute names

Nathan-Fenner:
A serieslist produced by evaluation now has a proper name, based on the functions which evaluated it.

For example, `select aggregate.sum(cpu group by 'app')` will have a name like `"aggregate.sum(cpu group by app)"`

@jeeyoungk @achow

### DIFF
--- a/query/function.go
+++ b/query/function.go
@@ -108,6 +108,7 @@ func MakeOperatorMetricFunction(op string, operator func(float64, float64) float
 			return seriesListValue(api.SeriesList{
 				Series:    result,
 				Timerange: context.Timerange,
+				Name:      fmt.Sprintf("(%s %s %s)", leftValue.name(), op, rightValue.name()),
 			}), nil
 		},
 	}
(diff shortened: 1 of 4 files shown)


query/value.go
```diff
@@ -79,14 +86,15 @@ func (value scalarValue) toSeriesList(timerange api.Timerange) (api.SeriesList,
...
-
 func (value scalarValue) toString() (string, error) {
 	return "", conversionError{"scalar", "string"}
 }
-
 func (value scalarValue) toScalar() (float64, error) {
 	return float64(value), nil
 }
+func (value scalarValue) name() string {
+	return fmt.Sprintf("%d", value)
```
jeeyoungk: `%f`?

jeeyoungk:
LGTM

jeeyoungk:
can we also add `transform.alias` ?

Nathan-Fenner:
alias function added

achow:
Looks like we might not be checking SeriesList.Name in our tests. We probably should (or otherwise pull out the renaming logic into their own functions for testing).

jeeyoungk:
@achow - you're right - we should add tests for them as a separate tests in `command_test.go`

query/function.go
```diff
@@ -163,13 +176,18 @@ func MakeTransformMetricFunction(name string, parameterCount int, transformer fu
 			if err != nil {
 				return nil, err
 			}
+			parameterNames := ""
+			for _, param := range parameters {
```
jeeyoungk: it's probably better to store them in an array and run `Join()` on them

query/function.go
```diff
@@ -141,7 +153,6 @@ func MakeTransformMetricFunction(name string, parameterCount int, transformer fu
 		Name:        name,
 		MinArgument: parameterCount + 1,
 		MaxArgument: parameterCount + 1,
-		Groups:      true,
```
Nathan-Fenner: This was accidentally added to `Transform` instead of `Aggregate`, making it possible to use `group by` in transforms but not in aggregations.

PR #101, opened 2015-06-18
Change parsing rules to prevent syntax errors

Nathan-Fenner:
The basic strategy is as follows:

"Non-terminals" (which are roughly any rule in all-caps or a string literal) will be preceded (whenever they expect one) by an `_` whose rule is still just `SPACE*`.

Any keyword, where appearing as a string literal (e.g., `"select"` or `"where"`) must be followed by the `KEY` rule, which is just `!ID_CONT` (or `![a-zA-Z0-9]`).

For example:

```describeMetrics <- _ "metrics" KEY _ "where" KEY tagName _ "=" literalString { p.makeDescribeMetrics() }```

Since we have a keyword `"metrics"`, it is followed by `KEY`. However, it needs to take care of spaces that may precede it. It doesn't have to take care of spaces that appear after it. Similarly consider the `"="` which has a `_` preceding it. Since it's not a keyword (it's nonalphanumeric) there's no `KEY` following it.

There is no longer any "mandatory space" (the `KEY` declarations appear instead). Everyone is responsible for cleaning up spaces *before* them, but not *after*.

@jeeyoungk @achow 

Tests have been added.

### DIFF
--- a/query/language.peg
+++ b/query/language.peg
@@ -45,37 +45,36 @@ type Parser Peg {
 # Hierarchical Syntax
 # ===================
 
-root <- (selectStmt / describeStmt) !.
+root <- (selectStmt / describeStmt) _ !.
 
-selectStmt <- _ "select" __
+selectStmt <- _ "select" KEY
   expressionList
   optionalPredicateClause
   propertyClause {
     p.makeSelect()
   }
-  _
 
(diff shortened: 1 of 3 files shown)


jeeyoungk:
LGTM! thanks for fixing this.

PR #102, opened 2015-06-18
Implementing the concept of timeout.

jeeyoungk:
* Refactoring the backend so that it's easier to pass values between
single and multi backend
* implementing timeouts for the backend.
* moving main functions to separate packages.
* updating tests


api/backend/multibackend.go
```diff
@@ -69,32 +63,48 @@ func NewParallelMultiBackend(backend api.Backend, limit int) api.MultiBackend {
...
-		if thisErr != nil {
-			err = thisErr
+	for _ = range works {
+		select {
+		case thisErr := <-channel:
+			if thisErr != nil {
+				err = thisErr
+			}
+		case <-cancellable.Done():
+			err = api.BackendError{
```
Nathan-Fenner: It makes more sense to just return immediately, since it will match (nondeterministically) every remaining iteration of the loop.

jeeyoungk: you're right - i blindly copied the code.

Nathan-Fenner:
LGTM

PR #105, opened 2015-06-19
Define transform.nan_keep_last

Nathan-Fenner:
The function `transform.nan_keep_last` replaces any `NaN` with the first non-`NaN` that came before (unless there are none, in which case it stays `NaN`).


syamp:
Probably name it as fill_nan_last or something ?

achow:
Graphite uses keepLastValue. Maybe "nan_keep_last"?

jeeyoungk:
@achow @syamp bikeshedding the name! :)

jeeyoungk:
LGTM

achow:
LGTM

PR #108, opened 2015-06-19
Fixing Filter + Filter Tests

Nathan-Fenner:
The main issues with filtering are due to (1) inconsistent/confusing behavior for totally missing series (all NaN) and (2) the aggregating functions failed whenever a single entry in their argument arrays was NaN.

This PR moves the not-NaN filter from the `applyAggregate` function into each one of the aggregators, so that they can be used as useful summarizes for `FilterBy`.

The tests don't cover expression evaluation directly but do cover the functionality of `FilterBy()` which is used by expressions registered as filters.

`FilterBy`'s argument "`ascending`" has been renamed to the more-clear "`lowest`" meaning that when `true`, it will return the `N` lowest results (if false, it returns the `N` highest). [This change doesn't effect behavior in calling the function].

### DIFF
--- a/query/aggregate/aggregate.go
+++ b/query/aggregate/aggregate.go
@@ -84,7 +84,7 @@ func filterNaN(array []float64) []float64 {
 
 // The sum aggregator returns the mean of the given array
 func AggregateSum(array []float64) float64 {
-	// array = filterNaN(array)
+	array = filterNaN(array)
 	sum := 0.0
 	for _, v := range array {
 		sum += v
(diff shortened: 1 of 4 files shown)


query/aggregate/aggregate.go
```diff
@@ -72,8 +72,19 @@ func groupBy(list api.SeriesList, tags []string) []group {
...
+		if !math.IsNaN(v) {
+			result = append(result, v)
+		}
+	}
+	return result
+}
+
 // The sum aggregator returns the mean of the given array
 func AggregateSum(array []float64) float64 {
+	// array = filterNaN(array)
```
jeeyoungk: uncomment?

jeeyoungk:
LGTM

PR #109, opened 2015-06-20
Updating the front-end

jeeyoungk:
Updating the front-end.
    
    * extracted out the google chart loading.
    * tab -> space for consistency.
    * using queries instead of hashes. using HTML5 mode in angular, using
    $location instead of depending on window.location.
    * some change to the in-flight query detection.
    * some change to the label generation logic to use the `name` field.

### DIFF
--- a/main/static/index.html
+++ b/main/static/index.html
@@ -1,34 +1,35 @@
 <!doctype html>
 <html>
 <head>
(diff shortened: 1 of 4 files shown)


Nathan-Fenner:
LGTM

PR #111, opened 2015-06-20
Language support for "pipeline" style function application:

jeeyoungk:
Adds a new syntax `expr | func(...args group by y)`, which is
a different way of writing `func(expr, ...args group by y`

```
select
  `jvm.thread-count` | aggregate.max | transform.moving_average('2h')
where
  type = 'thread_count' and
  app in ('esperanto', 'eero', 'ledger')
  from -24h to now resolution 5m
```

### DIFF
--- a/query/language.peg
+++ b/query/language.peg
@@ -80,10 +80,13 @@ optionalPredicateClause <-
 
 expressionList <-
   { p.addExpressionList() }
-  expression_1 { p.appendExpression() }
+  expression_start{ p.appendExpression() }
   (
-    _ COMMA expression_1 { p.appendExpression() }
+    _ COMMA expression_start { p.appendExpression() }
   )*
+
+expression_start <-
+  expression_1
 
 expression_1 <-
   expression_2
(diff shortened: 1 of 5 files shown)


Nathan-Fenner:
I noticed that I missed some things when I changed around how whitespace was consumed. I don't think they'll cause any problems right now but for consistency's sake I think they should be fixed:

`COLUMN_NAME`, `TAG_NAME`, `METRIC_NAME`, `IDENTIFIER`, and `ID_SEGMENT` are all used in a capturing manner as in `<COLUMN_NAME>` and therefore none of them should consume spaces:

https://github.com/square/metrics/pull/111/files?diff=split#diff-cccc438327926daa1a0aecdb5d8ef353L203

Nathan-Fenner:
LGTM

query/language.peg
```diff
@@ -150,7 +175,7 @@ predicateClause <- _ "where" KEY _ predicate_1
 
 predicate_1 <-
   predicate_2 _ OP_OR predicate_1 { p.addOrPredicate() } /
-  predicate_2 /
```
jeeyoungk: this was a bug.

achow:
Should add a pipe integration test, but otherwise LGTM.

PR #112, opened 2015-06-22
Collect data in Profilers stored in Contexts

Nathan-Fenner:
`EvaluationContext` and `Fetch*Request` are given a `Profiler` field of type `*inspect.Profiler` which collects metrics about the evaluation of each command.

The interface for the `Profiler` is a method `func (p *Profiler) Add(name string) func()` which returns a function that adds a profiling metric from the whose `start` is the time of the first call, and whose `finish` is the time of the returned function's call, storing it in `name`.

Typical usage is: 
```
defer request.Profiler.Add("FetchSingleRequest")
```

There is no need to check for a `nil` pointer before doing this call. If the pointer is `nil`, the `.Add` method has no effect.

If it's necessary to time a function that is invoked (rather than modifying the function to do the profiling itself), then you can do the following (for example);

```
toGraphiteNameDone := request.Profiler.Add("api.ToGraphiteName")
graphiteName, err := request.API.ToGraphiteName(request.Metric)
toGraphiteNameDone()
```

However, Backends and APIs do not need to invoke these. Instead, use the wrappers `api.ProfilingAPI`, `api.ProfilingBackend`, `api.ProfilingMultiBackend`, and `query.ProfilingCommand`.

TODO: tests

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -18,6 +18,8 @@
 // for the terminology.
 package api
 
+import "github.com/square/metrics/inspect"
+
 // API is the set of public methods exposed by the indexer library.
 type API interface {
 	// AddMetric adds the metric to the system.
@@ -43,6 +45,41 @@ type API interface {
 	// For a given tag key-value pair, obtain the list of all the MetricKeys
 	// associated with them.
 	GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error)
+}
+
+// ProfilingAPI wraps an ordinary API and also records profiling metrics to a given Profiler object.
+type ProfilingAPI struct {
+	Profiler *inspect.Profiler
+	API      API
+}
+
+func (api ProfilingAPI) AddMetric(metric TaggedMetric) error {
+	defer api.Profiler.Add("api.AddMetric")
+	return api.API.AddMetric(metric)
+}
+func (api ProfilingAPI) RemoveMetric(metric TaggedMetric) error {
+	defer api.Profiler.Add("api.RemoveMetric")
+	return api.API.RemoveMetric(metric)
+}
+func (api ProfilingAPI) ToGraphiteName(metric TaggedMetric) (GraphiteMetric, error) {
+	defer api.Profiler.Add("api.ToGraphiteName")
+	return api.API.ToGraphiteName(metric)
+}
+func (api ProfilingAPI) ToTaggedName(metric GraphiteMetric) (TaggedMetric, error) {
+	defer api.Profiler.Add("api.ToTaggedName")
+	return api.API.ToTaggedName(metric)
+}
+func (api ProfilingAPI) GetAllTags(metricKey MetricKey) ([]TagSet, error) {
+	defer api.Profiler.Add("api.GetAllTags")
+	return api.API.GetAllTags(metricKey)
+}
+func (api ProfilingAPI) GetAllMetrics() ([]MetricKey, error) {
+	defer api.Profiler.Add("api.GetAllMetrics")
+	return api.API.GetAllMetrics()
+}
+func (api ProfilingAPI) GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error) {
+	defer api.Profiler.Add("api.GetMetircsForTag")
+	return api.API.GetMetricsForTag(tagKey, tagValue)
 }
 
 // Configuration is the struct that tells how to instantiate a new copy of an API.
--- a/api/api.go
+++ b/api/api.go
@@ -47,41 +47,6 @@ type API interface {
 	GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error)
 }
 
-// ProfilingAPI wraps an ordinary API and also records profiling metrics to a given Profiler object.
-type ProfilingAPI struct {
-	Profiler *inspect.Profiler
-	API      API
-}
-
-func (api ProfilingAPI) AddMetric(metric TaggedMetric) error {
-	defer api.Profiler.Add("api.AddMetric")
-	return api.API.AddMetric(metric)
-}
-func (api ProfilingAPI) RemoveMetric(metric TaggedMetric) error {
-	defer api.Profiler.Add("api.RemoveMetric")
-	return api.API.RemoveMetric(metric)
-}
-func (api ProfilingAPI) ToGraphiteName(metric TaggedMetric) (GraphiteMetric, error) {
-	defer api.Profiler.Add("api.ToGraphiteName")
-	return api.API.ToGraphiteName(metric)
-}
-func (api ProfilingAPI) ToTaggedName(metric GraphiteMetric) (TaggedMetric, error) {
-	defer api.Profiler.Add("api.ToTaggedName")
-	return api.API.ToTaggedName(metric)
-}
-func (api ProfilingAPI) GetAllTags(metricKey MetricKey) ([]TagSet, error) {
-	defer api.Profiler.Add("api.GetAllTags")
-	return api.API.GetAllTags(metricKey)
-}
-func (api ProfilingAPI) GetAllMetrics() ([]MetricKey, error) {
-	defer api.Profiler.Add("api.GetAllMetrics")
-	return api.API.GetAllMetrics()
-}
-func (api ProfilingAPI) GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error) {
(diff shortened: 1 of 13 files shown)


query/expression.go
```diff
@@ -116,6 +118,8 @@ func (expr stringExpression) Evaluate(context EvaluationContext) (value, error)
 }
 
 func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, error) {
+	// Record the duration of this request in the profiler.
+	defer context.Profiler.Add("metricFetchExpression.Evaluate()", inspect.Now())
```
jeeyoungk: does this work?, i thought `defer func()` gets called at the very end.

Nathan-Fenner: from http://blog.golang.org/defer-panic-and-recover

"1. A deferred function's arguments are evaluated when the defer statement is evaluated."

The function itself is called at the end, but the `inspect.Now()` is called here.

jeeyoungk: oh wow, that's ... subtle :P

jeeyoungk: another way to do this is i think, making an `Add()` function that returns a function. this will remove the need to have `inspect.Now()` as a parameter.

`defer context.Profile.Add("foo") -> returns func()`

Nathan-Fenner: Yeah, that's a good idea!

api/backend/multibackend.go
```diff
@@ -114,6 +117,9 @@ func (m *parallelMultiBackend) fetchManyLazy(cancellable api.Cancellable, works
 }
 
 func (m *parallelMultiBackend) FetchMultipleSeries(request api.FetchMultipleRequest) (api.SeriesList, error) {
+	// Both specific and general metrics (for multibackends in general, and parallel in particular) are produced.
```
jeeyoungk: let's remove paren in all the names - i think it's a bit more consistent.

inspect/profile.go
```diff
@@ -0,0 +1,71 @@
...
+import "time"
+
+// Profiler contains a sequence of profiles which are collected over the course of a query execution.
+type Profiler struct {
+	profiles []Profile
+}
+
+// Add will create a profile of the given name from `start` until the current time.
+// Add acts in a threadsafe manner.
+func (p *Profiler) Add(name string) func() {
```
jeeyoungk: let's have an example code.

inspect/profile.go
```diff
@@ -0,0 +1,71 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package inspect
+
+import "time"
+
+// Profiler contains a sequence of profiles which are collected over the course of a query execution.
+type Profiler struct {
+	profiles []Profile
```
jeeyoungk: so if we want to test this particular logic (regarding adding profiles, etc), we can do it by mocking out `time.Now()` somehow. there's two ways to do it in go.

1. make `time.Now()` as a member function of `Profiler`,
2. make `now` function package-private, and allow tests to override it.

inspect/profile.go
```diff
@@ -0,0 +1,71 @@
...
+// Add will create a profile of the given name from `start` until the current time.
+// Add acts in a threadsafe manner.
+func (p *Profiler) Add(name string) func() {
+	if p == nil {
+		// If the profiler instance doesn't exist, then don't attempt to operate on it.
+		return func() {}
+	}
+	start := time.Now()
+	return func() {
+		p.profiles = append(p.profiles, Profile{name: name, startTime: start, finishTime: time.Now()})
```
jeeyoungk: let's make it threadsafe (mutex)/

query/expression.go
```diff
@@ -152,6 +159,7 @@ func (expr *metricFetchExpression) Evaluate(context EvaluationContext) (value, e
 			context.Timerange,
 			context.API,
```
jeeyoungk: if we pass in `WrappedAPI{context.API, context.Profiler}`

inspect/profile.go
```diff
@@ -0,0 +1,165 @@
...
+	Profiler *Profiler
+	Command  query.Command
+}
+
+func (cmd ProfilingCommand) Name() string {
+	return cmd.Command.Name()
+}
+
+func (cmd ProfilingCommand) Execute(context query.ExecutionContext) (interface{}, error) {
+	defer cmd.Profiler.Add(fmt.Sprintf("%sCommand.Execute", cmd.Name()))
```
jeeyoungk:
```
api = profilingapi(originalapi)
...
```

main/ui/ui.go
```diff
@@ -30,9 +31,21 @@ func main() {
...
+	apiInstance := inspect.ProfilingAPI{
+		API:      common.NewAPI(config.API),
+		Profiler: profiler,
+	}
+	blueflood := inspect.ProfilingBackend{
+		Profiler: profiler,
+		Backend:  blueflood.NewBlueflood(config.Blueflood),
+	}
+	backend := inspect.ProfilingMultiBackend{
+		MultiBackend: backend.NewSequentialMultiBackend(blueflood),
```
jeeyoungk: we can make the integrator to provide `func (Backend) MultiBackend` instead of `Multibackend` as an input.

api/api.go
```diff
@@ -55,3 +57,38 @@ type Config struct {
 	Hosts    []string `yaml:"hosts"`
 	Keyspace string   `yaml:"keyspace"`
 }
+
+// ProfilingAPI wraps an ordinary API and also records profiling metrics to a given Profiler object.
+type ProfilingAPI struct {
```
jeeyoungk: should this go to `inspect`? I think I like the idea of `square/metrics/api` being a package without any dependency.

jeeyoungk: actually, nvm, `backend` has reference to `profiler`, which would not work currently.

inspect/profile.go
```diff
@@ -0,0 +1,87 @@
...
+		mutex:    &sync.Mutex{},
+		profiles: []Profile{},
+	}
+	return &profiler
+}
+
+// Record will create a profile of the given name from `start` until the current time.
+// Record acts in a threadsafe manner.
+func (p *Profiler) Record(name string) func() {
+	if p == nil {
```
jeeyoungk: such a fun hack.

ui/ui.go
```diff
@@ -78,11 +80,17 @@ func (q QueryHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reques
...
 
+	cmd, profiler := query.NewProfilingCommand(cmd)
 	result, err := cmd.Execute(q.context)
 	if err != nil {
 		errorResponse(writer, http.StatusInternalServerError, err)
 		return
 	}
+
+	// profiler holds all the profiling data from this execution.
+
```
jeeyoungk: nit-newline.

ui/ui.go
```diff
@@ -99,6 +107,8 @@ func (h StaticHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reque
 }
 
 func NewMux(config Config, context query.ExecutionContext) *http.ServeMux {
+	// Wrap the given API and Backend in their Profiling counterparts.
```
jeeyoungk: is this still relevant?

jeeyoungk:
LGTM (sans test).

I think few high-level integration test might be useful. do few queries.

* `describe all`
* `select ... where`

and check whether it has the expected # of statistics. (1 api call, 5 backend calls, etc). i think validating the actual data is unnecessary.

ui/ui.go
Nathan-Fenner: No it isn't.

jeeyoungk:
LGTM. Can you squash it?

query/inspect_test.go
```diff
@@ -0,0 +1,210 @@
...
+			expected: map[string]int{
+				"describe all.Execute": 1,
+				"api.GetAllMetrics":    1,
+			},
+		},
+	}
+
+	for _, test := range testCases {
+		cmd, err := Parse(test.query)
+		if err != nil {
```
jeeyoungk: probably want to `continue` here (since cmd is going to be `nil`)

query/inspect_test.go
```diff
@@ -0,0 +1,210 @@
+// Copyright 2015 Square Inc.
```
jeeyoungk: this file is nice.

jeeyoungk:
LGTM

PR #118, opened 2015-06-23
adding profiler.flush

jeeyoungk:
Adding the `Flush()` method - this is used to ensure that we can use `ProfilingAPI` in square's codebase to profile `API` that is not part of command - i.e. kafka consumers inserting metrics.


Nathan-Fenner:
LGTM

PR #119, opened 2015-06-24
Part 1 of extracting function out of `query`

jeeyoungk:
This is part 1 - I got sick of doing this work - so i'm putting it here.

value, expression, and EvaluationContext are all pulled out to `function` class - it mostly contains types & aggregate functions that are used by functions. Next todo items are:

* pulling transformations out (not hard, but tests have some dependency on nodes).
* pulling function registration out.


Nathan-Fenner:
LGTM

PR #120, opened 2015-06-24
Batch API requests

Nathan-Fenner:
@jeeyoungk @achow


jeeyoungk:
can we have a test adding multiple metrics?

achow:
Looks like this requires callers to batch instead of the API doing the batching? I think it'd be nicer if we did the latter.

PR #125, opened 2015-06-25
Adding profiling UI support

jeeyoungk:
Add a support for profiler.
    
    * added UI to expose the profiler.
    * added server-side support.
    * abstracted out the google chart integration a bit.
    * fixed the logger - using glog by default, passing varargs correctly, etc.
    * reduced a lot of logging levels to DEBUG


Nathan-Fenner:
LGTM

PR #132, opened 2015-06-29
Metric Sets / Caching Describe All

Nathan-Fenner:
TODO: verify functionality

### DIFF
--- a/internal/cassandra.go
+++ b/internal/cassandra.go
@@ -95,8 +95,8 @@ func (db *defaultDatabase) AddMetricName(metricKey api.MetricKey, tagSet api.Tag
 	}
 	db.allMetricsMutex.Lock()
 	// Remember the cached value so that it won't be written again in the absence of reads.
-	db.allMetricsMutex.Unlock()
 	db.allMetricsCache[metricKey] = true
+	db.allMetricsMutex.Unlock()
 	return nil
 
 }
--- a/internal/cassandra.go
+++ b/internal/cassandra.go
@@ -16,6 +16,8 @@
 package internal
 
 import (
+	"sync"
+
 	"github.com/gocql/gocql"
 	"github.com/square/metrics/api"
 )
@@ -42,6 +44,7 @@ type Database interface {
 type defaultDatabase struct {
 	session         *gocql.Session
 	allMetricsCache map[api.MetricKey]bool
+	allMetricsMutex *sync.RWMutex
 }
 
 // NewCassandraDatabase creates an instance of database, backed by Cassandra.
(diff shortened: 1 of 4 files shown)


internal/cassandra_test.go
```diff
@@ -39,6 +39,10 @@ func newDatabase(t *testing.T) *defaultDatabase {
 		t.Errorf("Cannot truncate")
 		return nil
 	}
+	if session.Query("TRUNCATE metric_name_set").Exec() != nil {
```
jeeyoungk: we should make this into a loop
```
for name := {...} {
  TRUNCATE!
}
```

internal/cassandra.go
```diff
@@ -40,7 +40,8 @@ type Database interface {
 }
 
 type defaultDatabase struct {
-	session *gocql.Session
+	session         *gocql.Session
+	allMetricsCache map[api.MetricKey]bool
```
jeeyoungk: you need to protect it with a mutex.

internal/cassandra.go
```diff
@@ -49,25 +61,67 @@ func NewCassandraDatabase(clusterConfig *gocql.ClusterConfig) (Database, error)
...
+		// If the key is found in the cache, exit early.
+		return nil
+	}
+	db.allMetricsMutex.RUnlock()
+	if err := db.session.Query("UPDATE metric_name_set SET metric_names = metric_names + ? WHERE shard = ?", []string{string(metricKey)}, 0).Exec(); err != nil {
+		return err
+	}
+	db.allMetricsMutex.Lock()
+	// Remember the cached value so that it won't be written again in the absence of reads.
+	db.allMetricsMutex.Unlock()
```
jeeyoungk: i think we got the order wrong.

internal/cassandra.go
```diff
@@ -49,25 +61,67 @@ func NewCassandraDatabase(clusterConfig *gocql.ClusterConfig) (Database, error)
...
+	return &defaultDatabase{
+		session:         session,
+		allMetricsCache: make(map[api.MetricKey]bool),
+		allMetricsMutex: &sync.RWMutex{},
+		tagIndexCache:   make(map[tagIndexCacheKey]bool),
+		tagIndexMutex:   &sync.RWMutex{},
+	}, nil
+}
+
+func eitherError(a error, b error) error {
```
jeeyoungk: i don't think we use this anymore?

internal/cassandra.go
```diff
@@ -107,18 +161,24 @@ func (db *defaultDatabase) GetMetricKeys(tagKey string, tagValue string) ([]api.
...
-		keys = append(keys, api.MetricKey(metricKey))
-	}
-	if err := iterator.Close(); err != nil {
+	err := db.session.Query("SELECT metric_names FROM metric_name_set WHERE shard = ?", 0).Scan(&keys)
+	if err != nil {
 		return nil, err
 	}
+	db.allMetricsMutex.Lock()
+	for _, key := range keys {
+		db.allMetricsCache[key] = true
```
jeeyoungk: should we do `db.allMetricsCache = empty map` before filling this in?

PR #135, opened 2015-06-30
Create and use duration type for values

Nathan-Fenner:
The `DurationValue` type represents quantities that are durations. There is no implicit conversion allowed between numbers and duration values.

@jeeyoungk @achow

### DIFF
--- a/function/value.go
+++ b/function/value.go
@@ -28,6 +28,7 @@ type Value interface {
 	ToSeriesList(api.Timerange) (api.SeriesList, error)
 	ToString() (string, error)
 	ToScalar() (float64, error)
+	ToDuration() (int64, error)
 	GetName() string
 }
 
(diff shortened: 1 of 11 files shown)


query/query_test.go
```diff
@@ -134,6 +134,8 @@ var selects = []string{
 	"x|y from 0 to 0",
 	"x|f + y*z from 0 to 0",
 	"x|f + y|g from 0 to 0",
+	"x|f + y|g(4) from 0 to 0",
```
jeeyoungk: we should add test for duration nodes.

function/value.go
```diff
@@ -28,6 +28,7 @@ type Value interface {
 	ToSeriesList(api.Timerange) (api.SeriesList, error)
 	ToString() (string, error)
 	ToScalar() (float64, error)
+	ToDuration() (int64, error)
```
jeeyoungk: can we make it return `time.Duration`?

jeeyoungk:
LGTM

PR #136, opened 2015-06-30
Embed Graphs

Nathan-Fenner:
The page `embed.html` will render a graph that fills the screen, using the URL parameters as the query and settings.

An `<iframe>` can link to the embed page for an MQE graph.

The JS was refactored to allow reduced duplication.

Autocom no longer interferes with Angular reading the DOM, so it's been un-hacked.

Autocom configuration parameters have been adjusted slightly.

### DIFF
--- a/main/static/script.js
+++ b/main/static/script.js
@@ -294,8 +294,6 @@ module.controller("mainCtrl", function(
 
   // Triggers when the button is clicked.
   $scope.onSubmitQuery = function() {
-    // TODO - unhack this.
-    $inputModel.query = document.getElementById("query-input").value;
     $location.search("query", $inputModel.query);
     $location.search("renderType", $inputModel.renderType);
     $location.search("profile", $inputModel.profile.toString());
--- a/main/static/script.js
+++ b/main/static/script.js
@@ -21,7 +21,7 @@ module.config(function($locationProvider) {
   $locationProvider.html5Mode(true);
 });
 
-module.service("$google", function($rootScope) {
+module.factory("$google", function($rootScope) {
   // abstraction over the async loading of google libraries.
   // registered functions are either invoked immediately (if the library finished loading).
   // or queued in an array.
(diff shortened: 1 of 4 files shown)


main/static/script.js
```diff
@@ -94,107 +74,98 @@ module.controller("mainCtrl", function(
...
+// chart-related functions
+function clearChart(chart) {
+  if (chart.chart !== null) {
+    chart.chart.clearChart();
+    chart.chart = null;
+  }
+};
+
+module.factory("$asyncRender", function($chartWaiting) {
+  return function($scope, chart, dataTable, options) {
```
jeeyoungk: should just inject `$rootScope` and run `$rootScope.$apply` - that's a bit simpler.

main/static/script.js
```diff
@@ -94,107 +74,98 @@ module.controller("mainCtrl", function(
...
+module.factory("$launchRequest", function($google, $http, $receive, $queryTicketBooth, $launchedQueries) {
+  return function($scope, params) {
+    var ticket = $queryTicketBooth.next();
+    $launchedQueries.inc();
     var request = $http.get("/query", {
       params:params
     }).success(function(data, status, headers, config) {
-      $scope.launchedQuery--;
+      $launchedQueries.dec();
       $scope.queryResult = data;
```
jeeyoungk: hmm. it's a bit weird to do `$scope.queryResult = ...` here. I think a better approach is for this function to return a promise, and the calling side to set the scope variable instead via a callback.

main/static/script.js
```diff
@@ -227,15 +198,17 @@ module.controller("mainCtrl", function(
       options.isStacked = true;
       chart.chart = new google.visualization.AreaChart(chart.dom);
     }
-    asyncRender(chart, dataTable, options);
+    $asyncRender($scope, chart, dataTable, options);
   };
+});
 
-  function receiveProfile(object, chart) {
+module.factory("$receiveProfile", function($chartWaiting, $asyncRender) {
```
jeeyoungk: i think both receiveSelect and receiveProfile can be written to _not_ use `$scope` - it's hard to reason around, because `$scope` can contain (virtually) anything. can we pass the required logic as a variable? (also, $scope.apply can be abstracted out to rootscope.$apply and simply be injected).

Nathan-Fenner: I cleaned up `$scope` and `$rootScope` everywhere, so now only the `$main`/`$embed` controllers inject `$scope` and the `$google` and `$asyncRender` factory inject `$rootScope`, and they're never passed as parameters.

main/static/script.js
Nathan-Fenner: I used a `$q.defer()` promise to clean it up! Now the callers just do `.then(function(data) { $scope.queryResult = data; });`

jeeyoungk:
LGTM

PR #138, opened 2015-07-01
FilterRecent

Nathan-Fenner:
This PR defines a series of functions `filter.recent_*` which act similarly to `filter.*`.

Their purpose is to filter based on recent activity, rather than on total activity. For example:

`select filter.recent_highest_max( cpu, 5, 1d ) from -2w to now`

will show CPU usage for 5 hosts over the past 2 weeks, choosing these hosts by their activity only in the last day.

### DIFF
--- a/function/filter/filter.go
+++ b/function/filter/filter.go
@@ -17,6 +17,7 @@ package filter
 import (
 	"math"
 	"sort"
+	"time"
 
 	"github.com/square/metrics/api"
 )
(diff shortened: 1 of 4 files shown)


jeeyoungk:
nice. can we add some integration test for this?

Nathan-Fenner:
Some of the integration tests are possibly a little bit fragile.

In particular, to make testing easier, I'm relying on the fact that `filter.*` sorts the resulting lists, although this isn't exactly a documented feature (since externally, there's no good way to predict series ordering).

This is a bit odd since if you ask for more series than you have, then sorting *does not* occur. I'm not sure how this should be addressed, if at all.

jeeyoungk:
@Nathan-Fenner I think always sorting the result is actually not a bad behavior. it would make the result more consistent (both for testing and usability).

function/filter/filter.go
```diff
@@ -78,3 +79,37 @@ func FilterBy(list api.SeriesList, count int, summary func([]float64) float64, l
 		Timerange: list.Timerange,
 	}
 }
+
+func FilterRecentBy(list api.SeriesList, count int, summary func([]float64) float64, lowest bool, duration time.Duration) api.SeriesList {
+	if len(list.Series) < count {
```
jeeyoungk: let's always sort the result.

jeeyoungk:
LGTM with nit

PR #140, opened 2015-07-01
Fix bug in parsing duration literals

Nathan-Fenner:
Requiring `KEY` to follow the duration suffix is necessary to parse everything correctly.

### DIFF
--- a/query/language.peg
+++ b/query/language.peg
@@ -290,7 +290,7 @@ NUMBER_FRACTION <- "." [0-9]+
 NUMBER_INTEGER  <- "-"? NUMBER_NATURAL
 NUMBER_EXP      <- "e" ("+" / "-")? [0-9]+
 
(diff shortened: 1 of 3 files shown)


achow:
Needs tests for something that would've broken if this weren't there.

Nathan-Fenner:
It's a bit difficult, because it replaces one error with another one.

achow:
So this actually breaks something now?

Nathan-Fenner:
No, previously you'd see the message `"Programming error"`. Now you will see `"Syntax Error"` on a query like:

`cpu | transform.moving_average( 10q ) from 0 to 0`

(The invalid duration suffix `q`)

Nathan-Fenner:
I found a good spot, since `query_test.go` has some checks for `SyntaxError`s in particular.

jeeyoungk:
LGTM

PR #141, opened 2015-07-01
More Functions - Bounding transforms and counting aggregates

Nathan-Fenner:
* `aggregate.total` tells how many timeseries are in the serieslist (or group)
* `aggregate.count` tells us how many values aren't `NaN`
* `transform.bound(low, high)` lets us place bounds on functions (they'll be clamped to the edges)
* `transform.lower_bound(low)` lets us place a lower bound
* `transform.upper_bound(high)` lets us place a higher bound

Also fixes a bug in `assert.EqFloatArray` (sign of `delta` was being ignored, so some differences would have gone unnoticed. No tests failed after correcting this behavior).

### DIFF
--- a/assert/assert.go
+++ b/assert/assert.go
@@ -96,7 +96,7 @@ func (assert Assert) EqFloatArray(actual, expected []float64, epsilon float64) {
 			}
 		} else {
 			delta := actual[i] - expected[i]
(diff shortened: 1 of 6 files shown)


jeeyoungk:
LGTM

PR #145, opened 2015-07-02
Define 'collapse by' clause

Nathan-Fenner:
The `combine by` clause acts oppositely to the `group by` clause.

Performing `cpu | aggregate.sum( collapse by host )` will eliminate the `host` tag by ignoring the host tag while aggregating - so series which are otherwise the same but with different hosts will be aggregated.

TODO: *integration* tests for collapse

### DIFF
--- a/function/aggregate/aggregate.go
+++ b/function/aggregate/aggregate.go
@@ -30,28 +30,52 @@ type group struct {
 
 // If the given group will accept this given series (since it belongs to this group)
 // then groupAccept will return true.
-func groupAccepts(left api.TagSet, right api.TagSet, tags []string) bool {
-	for _, tag := range tags {
-		if left[tag] != right[tag] {
-			return false
+func groupAccepts(left api.TagSet, right api.TagSet, tags []string, combines bool) bool {
+	if combines {
+		tolerateDifferent := map[string]bool{}
+		for _, tag := range tags {
+			tolerateDifferent[tag] = true
+		}
+		for tag, leftValue := range left {
+			rightValue, ok := right[tag]
+			if ok && !tolerateDifferent[tag] && leftValue != rightValue {
+				return false
+			}
+		}
+	} else {
+		for _, tag := range tags {
+			if left[tag] != right[tag] {
+				return false
+			}
 		}
 	}
 	return true
 }
 
 // addToGroup adds the series to the corresponding bucket, possibly modifying the input `rows` and returning a new list.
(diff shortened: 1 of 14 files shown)


query/language.peg
```diff
@@ -163,10 +165,18 @@ groupByClause <-
   }
   (
     _ COMMA _ <COLUMN_NAME> {
-    p.appendGroupBy(unescapeLiteral(buffer[begin:end]))
+      p.appendGroupBy(unescapeLiteral(buffer[begin:end]))
     }
   )*
 
+combineByClause <-
```
syamp: collapseByClause?

function/expression.go
```diff
@@ -32,18 +32,23 @@ type Registry interface {
 	All() []string                             // all the registered functions
 }
 
+type Groups struct {
```
jeeyoungk: let's comment on what these fields do.

function/aggregate/aggregate.go
```diff
@@ -64,10 +88,11 @@ func addToGroup(rows []group, series api.Timeseries, tags []string) []group {
 }
 
 // groupBy groups the given SeriesList by tags, producing a list of lists (of type groupResult)
-func groupBy(list api.SeriesList, tags []string) []group {
+// if `combines` is true, then it groups on all other tags instead.
+func groupBy(list api.SeriesList, tags []string, combines bool) []group {
 	result := []group{}
 	for _, series := range list.Series {
-		result = addToGroup(result, series, tags)
+		result = addToGroup(result, series, tags, combines)
```
jeeyoungk: both `addToGroup`'s new tag generation and `groupAccepts` are starting to become unwieldy. we're using the boolean `combines` to do function dispatching at this point.

there should be a better way to organize these function. i feel like addToGroup should accept two functions (one to project one tag set to another grouped tagset), and don't pass in `tags` or `combines` at all. - if we have that, `groupAccept` can use that projection function to compare the tags also, simplifying `if combine...` sprinkled around everywhere.

Nathan-Fenner: I found a different way of organizing things, by changing `groupAccepts` to not need either `tags` or `combines`, and instead just operate on the tagsets.

There's still one `if collapses { ...` left behind, but there's only the one, and I manage to avoid having to pass functions as parameters.

jeeyoungk: +1

jeeyoungk:
LGTM

PR #147, opened 2015-07-08
Tag Functions (Drop / Set)

Nathan-Fenner:
Defines two new functions:

`tag.drop(serieslist, tag)` which will remove the tag from all series.

`tag.set(serieslist, tag, value)` which sets the tag for all series.

### DIFF
--- a/function/tag/tag.go
+++ b/function/tag/tag.go
@@ -16,15 +16,17 @@ package tag
 
 import (
 	"github.com/square/metrics/api"
+	"github.com/square/metrics/function"
 )
 
 // dropTagSeries returns a copy of the timeseries where the given `dropTag` has been removed from its TagSet.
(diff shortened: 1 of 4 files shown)


function/tag/tag.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/square/metrics/api"
+)
+
+// dropTagSeries returns a copy of the timeseries where the given `dropTag` has been removed from its TagSet.
+func dropTagSeries(series api.Timeseries, dropTag string) api.Timeseries {
+	tagSet := api.NewTagSet()
+	for tag, val := range series.TagSet {
+		tagSet[tag] = val
+	}
+	delete(tagSet, dropTag)
```
jeeyoungk: let's filter it in the `if` statement instead of calling delete.

jeeyoungk:
LGTM

PR #149, opened 2015-07-08
Assertion Error Fix

Nathan-Fenner:
`AssertionError`s won't show up when the user makes a syntax error in numbers or durations.

(They will instead receive a proper `SyntaxError` message instead of "Programming error")


jeeyoungk:
LGTM

PR #151, opened 2015-07-08
Tweak autocompleter to be easier to use

Nathan-Fenner:
Fixes existing undesirable behavior in Autocom.

* Add a feature to penalize long candidate words (so shorter candidates will be suggested first)
* Add a feature so that autocomplete suggestions won't appear in the middle of words before special characters


jeeyoungk:
LGTM

PR #152, opened 2015-07-09
Rework errors and durations

Nathan-Fenner:
Errors are made somewhat more structured.

* Instances of `errors.New()` are replaced by typed implementing the `ExecutionError` interface (which also has a `.TokenName()`).
* String => Duration implicit casts are no longer allowed. This causes errors to have closer to parsing, rather than execution, and makes the language more consistent. Strings are now only used to refer to tag values (or tags keys with `tag.drop`/`tag.set`) and absolute timestamps.

### DIFF
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -76,7 +76,7 @@ var MovingAverage = function.MetricFunction{
 
 		newContext := context
 		timerange := context.Timerange
-		newContext.Timerange, err = api.NewTimerange(timerange.Start()-int64(limit-1)*timerange.Resolution(), timerange.End(), timerange.Resolution())
+		newContext.Timerange, err = api.NewSnappedTimerange(timerange.Start()-int64(limit-1)*timerange.Resolution(), timerange.End(), timerange.Resolution())
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


function/transform/special.go
```diff
@@ -76,7 +76,7 @@ var MovingAverage = function.MetricFunction{
 
 		newContext := context
 		timerange := context.Timerange
-		newContext.Timerange, err = api.NewTimerange(timerange.Start()-int64(limit-1)*timerange.Resolution(), timerange.End(), timerange.Resolution())
+		newContext.Timerange, err = api.NewSnappedTimerange(timerange.Start()-int64(limit-1)*timerange.Resolution(), timerange.End(), timerange.Resolution())
```
Nathan-Fenner: There shouldn't be any changes caused here - but this is slightly safer, as `NewSnappedTimerange` raises errors in fewer cases

function/value.go
```diff
@@ -34,12 +34,17 @@ type Value interface {
...
-	from string
-	to   string
+	from  string
+	to    string
+	value interface{}
 }
 
 func (e conversionError) Error() string {
-	return fmt.Sprintf("cannot convert from type %s to type %s", e.from, e.to)
+	return fmt.Sprintf("cannot convert %+v (type %s) to type %s", e.value, e.from, e.to)
```
jeeyoungk: isn't `e.value` potentially super large? (if it's a `seriesList`).

Nathan-Fenner: Yeah, I could just change it to a string and add a `fmt.Sprintf("%f")` for the scalars.

jeeyoungk:
LGTM

PR #156, opened 2015-07-09
Select coerces results to type SeriesList

Nathan-Fenner:
This allows you to graph (for example):

`select cpu, 100 from -10m to now` to get a straight line at 100. Certain queries, like `select 'string' from -10m to now` will now error (failing to convert string => series list) instead of silently succeeding with a useless result.


jeeyoungk:
LGTM

PR #158, opened 2015-07-10
Short, tabular display for 'describe' queries

Nathan-Fenner:
Instead of listing every tagset for a given metric, the server computes a `map[string] []string` that it sends to the client, mapping each `key` to a list of associated `value`s.

The results are displayed in a series of tables (`display: inline-block; float: left`) with a small margin between them.

### DIFF
--- a/main/static/index.html
+++ b/main/static/index.html
@@ -71,9 +71,15 @@
 
     <div>
       <div class="col-xs-12">
-        <table ng-show="isTabular()" class="table table-striped">
+        <table ng-show="isTabular() && queryResult.name != 'describe'" class="table table-striped">
           <tr ng-repeat="row in queryResult.body"><td>{{row}}</td></tr>
         </table>
+        <div ng-show="isTabular() && queryResult.name == 'describe'">
+          <table ng-repeat="(key, list) in queryResult.body" class="table table-striped table-column">
+            <tr><th>{{key}}</th></tr>
+            <tr ng-repeat="value in list"><td>{{value}}</td></tr>
+          </table>
+        </div>
         <div>{{ (queryResult || "").trim() }}</div>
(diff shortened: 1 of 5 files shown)


main/static/script.js
```diff
@@ -367,8 +367,10 @@ module.controller("mainCtrl", function(
       return "rendered";
     }
   };
+  scope = $scope;
```
jeeyoungk: why do we need this?

main/static/index.html
```diff
@@ -71,9 +71,15 @@
 
     <div>
       <div class="col-xs-12">
-        <table ng-show="isTabular()" class="table table-striped">
+        <table ng-show="isTabular() && queryResult.name != 'describe'" class="table table-striped">
           <tr ng-repeat="row in queryResult.body"><td>{{row}}</td></tr>
         </table>
+        <div ng-show="isTabular() && queryResult.name == 'describe'">
+          <table ng-repeat="(key, list) in queryResult.body" class="table table-striped table-column">
```
jeeyoungk: can we transform the data, so that we can display them in

```
|A | B | C|
|---------|
|a1| b1| c1|
|a2| b2| c2|
|a3| b3| c3|
```
?

we can do the transformation on the client side.

main/static/script.js
Nathan-Fenner: oops, debugging I forgot about

main/static/index.html
Nathan-Fenner: It would be a bit strange. The columns aren't the same lengths, so you'd have a very, very jagged table in lots of cases (1 dc, 200 hosts!) I don't know whether this would be desirable.

jeeyoungk:
LGTM

PR #159, opened 2015-07-10
Add configurable limit on resolution slots

Nathan-Fenner:
Default: 1000 slots (8.3 hours at 30s resolution).

It can be configured through the `ExecutionContext`.

Note: this is only based on the original query for timeseries.

* Asking for many expressions: `select X, X, X, X, X, X, X, X, X, X, X, X, X, X, X, X, X, X from -8h to now` will select and return 17280 points
* `transform.moving_average` expands the search range, so `select transform.moving_average(X, 10d) from -1h to now` will select 28000 points of data, but only return the last 120 of these.

### DIFF
--- a/query/command.go
+++ b/query/command.go
@@ -33,6 +33,7 @@ type ExecutionContext struct {
 	Timeout    time.Duration     // optional
 	Profiler   *inspect.Profiler // optional
 	Registry   function.Registry // optional
+	SlotLimit  int               // optional (0 => default 1000)
 }
 
 // Command is the final result of the parsing.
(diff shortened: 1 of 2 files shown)


query/command.go
```diff
@@ -112,6 +113,21 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (interface{}, error)
...
+	}
+	if timerange.Slots() > slotLimit {
+		var limitMessage string
+		if context.SlotLimit == 0 {
+			limitMessage = "the default limit %d (no limit has been configured)"
+		} else {
+			limitMessage = "the configured limit %d"
+		}
+		limitMessage = fmt.Sprintf(limitMessage, slotLimit)
+		return nil, fmt.Errorf("Requested number of data points (%d) exceeds %s", timerange.Slots(), limitMessage)
```
jeeyoungk: we should throw an error similar to the error we get from the execution error.

Nathan-Fenner: Right now, the fetch-limit-error is also like this. Since they don't really have a "token" attached to them, I wasn't sure how to stuff them into that type...

jeeyoungk:
LGTM

PR #161, opened 2015-07-10
Natural Sort for Strings

Nathan-Fenner:
Defining this is a bit ridiculous. The point is to provide a better sorting experience for users. Lexicographical sorting (with ASCII order) isn't really what a human would expect. For example:

Apple < Zoo < apple < file100 < file2 < file90 < file99 < zoo

This package `natural_sort` defines a function `natural_sort.Sort(array []string)` that sorts its arguments in a natural order:

Apple < apple < file2 < file90 < file99 < file100 < Zoo < zoo

For reference, see:
https://en.wikipedia.org/wiki/Natural_sort_order
http://blog.codinghorror.com/sorting-for-humans-natural-sort-order/

### DIFF
--- a/query/natural_sort/natural_test.go
+++ b/query/natural_sort/natural_test.go
@@ -52,6 +52,7 @@ func TestRandom(t *testing.T) {
 	a.Eq(test, expected)
 	for i := 0; i < 1000; i++ {
 		testShuffle(test)
+		a := a.Contextf("input: %+v", test)
 		Sort(test)
 		a.Eq(test, expected)
 	}
--- a/query/natural_sort/natural_test.go
+++ b/query/natural_sort/natural_test.go
@@ -1,0 +1,58 @@
+// Copyright 2015 Square Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


query/natural_sort/natural_test.go
```diff
@@ -0,0 +1,58 @@
...
+func TestRandom(t *testing.T) {
+	a := assert.New(t)
+	expected := []string{"Apple", "apple", "file2", "file22", "file90", "file99", "file100", "Zoo", "zoo"}
+	test := []string{"Apple", "apple", "file2", "file22", "file90", "file99", "file100", "Zoo", "zoo"}
+	Sort(test)
+	a.Eq(test, expected)
+	for i := 0; i < 1000; i++ {
+		testShuffle(test)
+		Sort(test)
+		a.Eq(test, expected)
```
jeeyoungk: should do `a := a.Contextf(input:%s)` or something like that to print the randomized input, so that when the test fails we know why it fails.

achow:
fwiw, you can probably just sort on ToLower(), and it'd probably be fine to not really consider case variants since we'll never have them (and if we do, they won't be very useful).

jeeyoungk:
LGTM

PR #162, opened 2015-07-13
Jee.20150713.ui fixes

jeeyoungk:
* added an error type to handle resource-limiting behavior (timeout, fetches, slots, etc)
* removed main/ui/ui.go-specific configs from ui/ui.go
* changed the error to use a monospace font.
* adding `application/json` header.

### DIFF
--- a/function/expression.go
+++ b/function/expression.go
@@ -82,7 +82,7 @@ func (c FetchCounter) Limit() int {
 }
 
 func (c FetchCounter) Current() int {
-	return int(atomic.LoadInt32(c.count)) - c.limit
+	return c.limit - int(atomic.LoadInt32(c.count))
 }
 
 // Consume decrements the internal counter and returns whether the result is at least 0.
--- a/function/expression.go
+++ b/function/expression.go
@@ -65,13 +65,20 @@ func (f MetricFunction) Evaluate(context EvaluationContext,
 // fetchCounter is used to count the number of fetches remaining in a thread-safe manner.
 type FetchCounter struct {
 	count *int32
+	limit int
 }
 
 func NewFetchCounter(n int) FetchCounter {
 	n32 := int32(n)
 	return FetchCounter{
 		count: &n32,
+		limit: n,
 	}
+}
+
+// limit returns the max # of fetches allowed by this counter.
+func (c FetchCounter) Limit() int {
+	return c.limit
 }
 
 // Consume decrements the internal counter and returns whether the result is at least 0.
--- a/function/expression.go
+++ b/function/expression.go
@@ -81,6 +81,10 @@ func (c FetchCounter) Limit() int {
 	return c.limit
 }
 
+func (c FetchCounter) Current() int {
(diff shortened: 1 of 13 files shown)


query/command.go
```diff
@@ -143,7 +143,7 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (interface{}, error)
 			limitMessage = "the configured limit %d"
 		}
 		limitMessage = fmt.Sprintf(limitMessage, slotLimit)
-		return nil, fmt.Errorf("Requested number of data points (%d) exceeds %s", timerange.Slots(), limitMessage)
+		return nil, function.NewLimitError(fmt.Sprintf("Requested number of data points (%d) exceeds %s", timerange.Slots(), limitMessage), slotLimit)
```
achow: Interpolating `limitMessage` seems so awkward...

achow:
To try to to standardize the limit messages, maybe we can set a `context` and a `limit` and build Error() with something like `"#{context}: limit exceeded (#{limit})"`? So a message might be `"fetch timeout: limit exceeded (60s)"`.

If we're creating special error types, it should have some kind of standard format.

main/common/common.go
```diff
@@ -38,10 +38,16 @@ var (
 	Logger     = flag.String("logger", "glog", "Selects the logger to use")
 )
 
+type UIConfig struct {
+	Port      int `yaml:"port"`
```
achow: Why pull these fields out a level?

query/command.go
```diff
@@ -143,7 +143,9 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (interface{}, error)
 			limitMessage = "the configured limit %d"
 		}
 		limitMessage = fmt.Sprintf(limitMessage, slotLimit)
```
achow: Should clean up `limitMessage` if we don't plan on using it anymore.

function/expression.go
```diff
@@ -65,15 +65,26 @@ func (f MetricFunction) Evaluate(context EvaluationContext,
...
+		limit: n,
 	}
 }
 
+// limit returns the max # of fetches allowed by this counter.
+func (c FetchCounter) Limit() int {
+	return c.limit
+}
+
+func (c FetchCounter) Current() int {
```
achow: What is this supposed to do? Looks like it's only going to be negative.

main/common/common.go
jeeyoungk: there's two levels of configs.

1. running UI standalone (square/metrics) repo
2. running UI as part of an existing server (our internal repo).

these settings - port and timeout - are only required for (1) but the config was provided for both. i wanted to simplify the config from outside, thus split into two.

query/command.go
jeeyoungk: removed these lines.

function/expression.go
jeeyoungk: i fixed it. it was supposed to return the current # of fetches. added tests for it.

achow:
LGTM

PR #164, opened 2015-07-14
Implementing a parallel execute

jeeyoungk:
There are few cases where we want to evaluate subexpressions in
parallel. This PR abstracts out that logic so that it can be used in
different places, including:

* binary operators.
* top-level command evaluation.


Nathan-Fenner:
LGTM

PR #166, opened 2015-07-14
Adding "matches" statement

jeeyoungk:
grammatically, "describe all MATCHING ..." is more correct, but i'm a
bit conflicted - i don't want to have both MATCHES and MATCHING as
keywords.

WDYT?

### DIFF
--- a/query/parser.go
+++ b/query/parser.go
@@ -229,13 +229,9 @@ func (p *Parser) makeDescribe() {
 		p.flagTypeAssertion()
 		return
 	}
-	stringLiteral, ok := p.popNode(stringLiteralPointer).(*stringLiteral)
-	if !ok {
-		p.flagTypeAssertion()
-		return
-	}
+	literal := p.popStringLiteral()
 	p.command = &DescribeCommand{
-		metricName: api.MetricKey(stringLiteral.literal),
+		metricName: api.MetricKey(literal),
 		predicate:  predicateNode,
 	}
 }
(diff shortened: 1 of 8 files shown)


query/parser.go
```diff
@@ -263,24 +259,34 @@ func (p *Parser) makeSelect() {
 	}
 }
 
-func (p *Parser) makeDescribeAll() {
-	p.command = &DescribeAllCommand{}
+func (p *Parser) addNullMatchesClause() {
+	p.pushNode(&matcherClause{regexp.MustCompile("")})
```
achow: `matcherClause{regex:...}`

query/parser.go
```diff
@@ -263,24 +259,34 @@ func (p *Parser) makeSelect() {
...
 }
 
-func (p *Parser) makeDescribeMetrics() {
-	// Pop off the value.
-	stringLiteral, ok := p.popNode(stringLiteralPointer).(*stringLiteral)
+func (p *Parser) addMatchesClause() {
+	compiled := p.popRegex()
+	p.pushNode(&matcherClause{compiled})
```
achow: `matcherClause{regex:...}`

query/parser.go
```diff
@@ -263,24 +259,34 @@ func (p *Parser) makeSelect() {
...
+}
+
+func (p *Parser) makeDescribeAll() {
+	matcherClause, ok := p.popNode(matcherClausePointer).(*matcherClause)
 	if !ok {
 		p.flagTypeAssertion()
 		return
 	}
+	p.command = &DescribeAllCommand{matcherClause}
```
achow: `DescribeAllCommand{matcher:...}`

achow:
I think we have a few alternatives:

- Use `match`. It's more of an imperative, but I think it's kind of fits?
- Use `like`, like SQL.
- Cheese it with supporting pipes. We can `describe all | grep`.

Otherwise LGTM after nits.

jeeyoungk:
@achow

* nits fixed.
* matches -> match.

PTAL

Nathan-Fenner:
LGTM

syamp:
I think "like" is reads better.

PR #167, opened 2015-07-14
fixing parser syntax error panic

jeeyoungk:
if a syntax error occurs end-of-line, then it used to cause the parser to panic


Nathan-Fenner:
LGTM

PR #170, opened 2015-07-16
Refactoring resolution choosing code a bit

jeeyoungk:
* refactoring the resolution choosing code a bit.
* less hard-coding of resolution choosing code.
* renamed `Timerange.Resolution()` to `ResolutionMillis()`. added `Timerange.Resolution()` which returns the result in `time.Duration`.

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -35,21 +35,21 @@ type httpClient interface {
 }
 
 type Config struct {
-	BaseUrl  string               `yaml:"base_url"`
-	TenantId string               `yaml:"tenant_id"`
-	Ttls     map[Resolution]int64 `yaml:"ttls"` // Ttl in days
-	Timeout  time.Duration        `yaml:"timeout"`
+	BaseUrl  string           `yaml:"base_url"`
+	TenantId string           `yaml:"tenant_id"`
+	Ttls     map[string]int64 `yaml:"ttls"` // Ttl in days
+	Timeout  time.Duration    `yaml:"timeout"`
 }
 
-func (c Config) getTtlInMillis(r Resolution) int64 {
+func (c Config) getTTL(r Resolution) time.Duration {
 	var ttl int64
-	if v, ok := c.Ttls[r]; ok {
+	if v, ok := c.Ttls[r.bluefloodEnum]; ok {
 		ttl = v
 	} else {
 		// Use blueflood defaults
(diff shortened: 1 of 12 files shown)


api/backend/blueflood/blueflood.go
```diff
@@ -238,10 +259,6 @@ func addMetricPoint(metricPoint metricPoint, field func(metricPoint) float64, ti
 
 func bucketsFromMetricPoints(metricPoints []metricPoint, resultField func(metricPoint) float64, timerange api.Timerange) [][]float64 {
 	buckets := make([][]float64, timerange.Slots())
-	// Make the buckets:
```
jeeyoungk: no need - nil arrays can be appended in golang.

api/backend/blueflood/blueflood.go
```diff
@@ -119,8 +130,11 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
 	if !ok {
 		return api.Timeseries{}, fmt.Errorf("unsupported SampleMethod %s", request.SampleMethod.String())
 	}
+	queryResolution := b.config.bluefloodResolution(
+		time.Duration(request.Timerange.Resolution()),
```
Nathan-Fenner: The case the `time.Duration` should be unnecessary, since `Timerange.Resolution()` now returns a `time.Duration` object.

Nathan-Fenner: The `.Resolution()` method on `Timerange` objects should already be returning a `time.Duration` type, so the conversion can be omitted.

api/backend/blueflood/blueflood.go
```diff
@@ -286,21 +303,23 @@ var samplerMap map[api.SampleMethod]sampler = map[api.SampleMethod]sampler{
 
 // Blueflood keys the resolution param to a java enum, so we have to convert
 // between them.
-func (c Config) bluefloodResolution(resolution int64, timestamp int64) string {
+func (c Config) bluefloodResolution(
+	desiredResolution time.Duration,
+	startMs int64) Resolution {
 	now := time.Now().UTC().Unix() * 1000
```
Nathan-Fenner: Calling `.UTC()` has no effect here- it only changes the location of the time, which doesn't change the number of seconds since the Unix epoch.

Nathan-Fenner:
LGTM with little comments above

PR #171, opened 2015-07-16
Switch to config directory for rules

achow:
@jeeyoungk

### DIFF
--- a/internal/api.go
+++ b/internal/api.go
@@ -3,6 +3,8 @@ package internal
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"sort"
 	"time"
 
 	"github.com/gocql/gocql"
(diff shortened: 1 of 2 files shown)


internal/api.go
```diff
@@ -43,6 +37,38 @@ func NewAPI(config api.Config) (api.API, error) {
...
+	ruleSet := RuleSet{
+		rules: []Rule{},
+	}
+
+	filenames, err := filepath.Glob(filepath.Join(conversionRulesPath, "*.yaml"))
+	if err != nil {
+		return RuleSet{}, err
+	}
+
+	for _, filename := range filenames {
```
jeeyoungk: can we sort the filenames so that we can have a deterministic loading?

internal/api.go
```diff
@@ -43,6 +37,38 @@ func NewAPI(config api.Config) (api.API, error) {
...
+		file, err := os.Open(filename)
```
jeeyoungk: do we need to close files? (i've never done that in go so i have no clue whether that's needed or not).

Nathan-Fenner: The Go idiom is to have `defer file.Close();` immediately after opening a file (unless it needs to stay open after you return for some reason).

Nathan-Fenner: It's idiomatic to check for an error opening the file, and then defer closing the file.

For reference: https://gobyexample.com/writing-files

jeeyoungk:
LGTM

PR #173, opened 2015-07-17
Empty query does not cause panic

jeeyoungk:


achow:
LGTM

query/query_test.go
```diff
@@ -152,6 +152,7 @@ var selects = []string{
 
 // these queries should fail with a syntax error.
 var syntaxErrorQuery = []string{
+	"",
```
Nathan-Fenner: also add testcases for `" "`, `"\n"` and `" \n"`?

jeeyoungk: done

PR #176, opened 2015-07-22
Unmatched Graphite Rules

Nathan-Fenner:
This PR adds a way to query graphite metrics directly from MQE, even without rules for them. This is accomplished by use of the highly magical `graphite` function.

For example, the query `graphite("mqe.%host%.%dc%.cpu")` will produce a serieslist of every graphite name matching this pattern (like Graphite's `mqe.*.*.cpu`)

In order to accomplish this, various changes were made to API-related behavior. These changes are reverse-compatible with the following exception: the special metric name `"$graphite"` is assumed by the default API type to refer to a graphite metric name, rather than an MQE metric name.

In order to index graphite metrics, the API object must be cast to an `api.GraphiteStore` interface, and then the `.AddGraphiteMetric(metric) error` method should be called for each Graphite metric encountered. They're stored in a table with a TTL of about 8 hours.

TODO: integration tests?

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -62,8 +62,14 @@ type Config struct {
 	// Database configurations
 	// mostly cassandra configurations from
 	// https://github.com/gocql/gocql/blob/master/cluster.go
-	Hosts    []string `yaml:"hosts"`
-	Keyspace string   `yaml:"keyspace"`
+	Hosts    []string       `yaml:"hosts"`
+	Keyspace string         `yaml:"keyspace"`
+	Database DatabaseConfig `yaml:"database"`
+}
+
+// Configuration for the Database attached to the default API
+type DatabaseConfig struct {
+	GraphiteMetricTTL int `yaml:graphite_metric_ttl` // in seconds
 }
 
 // ProfilingAPI wraps an ordinary API and also records profiling metrics to a given Profiler object.
--- a/api/api.go
+++ b/api/api.go
@@ -56,6 +56,9 @@ type APIGraphiteStore interface {
 
 	// Get all the Graphite metrics
 	GetAllGraphiteMetrics() ([]GraphiteMetric, error)
+
+	// Checks whether (at runtime) the object actually is capable of performing these queries
+	SupportsGraphiteStore() bool
 }
 
 // Configuration is the struct that tells how to instantiate a new copy of an API.
@@ -118,4 +121,10 @@ func (api ProfilingAPI) GetAllGraphiteMetrics() ([]GraphiteMetric, error) {
 	} else {
 		return nil, nil
 	}
+}
+func (api ProfilingAPI) SupportsGraphiteStore() bool {
+	if apiGraphite, ok := api.API.(APIGraphiteStore); ok {
+		return apiGraphite.SupportsGraphiteStore()
+	}
+	return false
 }
--- a/api/api.go
+++ b/api/api.go
@@ -47,6 +47,17 @@ type API interface {
 	GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error)
 }
 
+type APIGraphiteStore interface {
+	// Embed the API interface
+	API
+
+	// Add a Graphite metric to the complete list (as needed)
+	AddGraphiteMetric(metric GraphiteMetric) error
+
+	// Get all the Graphite metrics
+	GetAllGraphiteMetrics() ([]GraphiteMetric, error)
+}
+
 // Configuration is the struct that tells how to instantiate a new copy of an API.
 type Config struct {
 	// Location of conversion rules. All *.yaml files in here will be loaded.
@@ -92,4 +103,19 @@ func (api ProfilingAPI) GetAllMetrics() ([]MetricKey, error) {
 func (api ProfilingAPI) GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error) {
 	defer api.Profiler.Record("api.GetMetricsForTag")()
 	return api.API.GetMetricsForTag(tagKey, tagValue)
+}
+func (api ProfilingAPI) AddGraphiteMetric(metric GraphiteMetric) error {
+	if apiGraphite, ok := api.API.(APIGraphiteStore); ok {
+		defer api.Profiler.Record("api.AddGraphiteMetric")()
+		return apiGraphite.AddGraphiteMetric(metric)
+	}
+	return nil
+}
+func (api ProfilingAPI) GetAllGraphiteMetrics() ([]GraphiteMetric, error) {
+	if apiGraphite, ok := api.API.(APIGraphiteStore); ok {
+		defer api.Profiler.Record("api.GetAllGraphiteMetrics")()
+		return apiGraphite.GetAllGraphiteMetrics()
+	} else {
+		return nil, nil
+	}
 }
--- a/api/api.go
+++ b/api/api.go
@@ -111,9 +111,9 @@ const SpecialGraphiteName = "$graphite"
 
 func (api ProfilingAPI) AddGraphiteMetric(metric GraphiteMetric) error {
 	defer api.Profiler.Record("api.AddGraphiteMetric")()
-	return api.AddGraphiteMetric(metric)
+	return api.API.AddGraphiteMetric(metric)
 }
 func (api ProfilingAPI) GetAllGraphiteMetrics() ([]GraphiteMetric, error) {
 	defer api.Profiler.Record("api.GetAllGraphiteMetrics")()
-	return api.GetAllGraphiteMetrics()
+	return api.API.GetAllGraphiteMetrics()
 }
--- a/api/api.go
+++ b/api/api.go
@@ -100,6 +100,9 @@ func (api ProfilingAPI) GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, 
 	defer api.Profiler.Record("api.GetMetricsForTag")()
 	return api.API.GetMetricsForTag(tagKey, tagValue)
 }
+
+const SpecialGraphiteName = "$graphite"
+
 func (api ProfilingAPI) AddGraphiteMetric(metric GraphiteMetric) error {
 	defer api.Profiler.Record("api.AddGraphiteMetric")()
 	return api.AddGraphiteMetric(metric)
--- a/api/api.go
+++ b/api/api.go
@@ -53,9 +53,6 @@ type GraphiteStore interface {
 
 	// Get all the Graphite metrics
 	GetAllGraphiteMetrics() ([]GraphiteMetric, error)
-
-	// Checks whether (at runtime) the object actually is capable of performing these queries
-	SupportsGraphiteStore() bool
 }
 
 // Configuration is the struct that tells how to instantiate a new copy of an API.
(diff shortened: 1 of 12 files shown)


function/graphite/graphite.go
```diff
@@ -0,0 +1,77 @@
...
+	}, true
+}
+
+func GetGraphiteMetrics(pattern string, API api.API) []api.TaggedMetric {
+	graphiteAPI, ok := API.(api.APIGraphiteStore)
+	if !ok || !graphiteAPI.SupportsGraphiteStore() {
+		// API is not able to fetch metrics
+		return nil
+	}
+	pieces := strings.Split(pattern, "%")
```
Nathan-Fenner: check that `len(pieces) % 2 == 1` so that you don't have invalid patterns

api/api.go
```diff
@@ -47,6 +47,20 @@ type API interface {
...
+	API
+
+	// Add a Graphite metric to the complete list (as needed)
+	AddGraphiteMetric(metric GraphiteMetric) error
+
+	// Get all the Graphite metrics
+	GetAllGraphiteMetrics() ([]GraphiteMetric, error)
+
+	// Checks whether (at runtime) the object actually is capable of performing these queries
+	SupportsGraphiteStore() bool
```
jeeyoungk: i think this feels a bit weird, since individual methods are capable of returning 'not supported error'.

function/registry/registry.go
```diff
@@ -76,6 +77,9 @@ func init() {
 	// Tags
 	MustRegister(tag.DropFunction)
 	MustRegister(tag.SetFunction)
+
+	// Really weird ones
+	MustRegister(graphiteSelect)
```
jeeyoungk: best comment ever.

internal/cassandra.go
```diff
@@ -41,6 +41,16 @@ type Database interface {
 	RemoveFromTagIndex(tagKey, tagValue string, metricKey api.MetricKey) error
 }
 
+type DatabaseGraphiteStore interface {
```
jeeyoungk: why do we have sudden hierarchy? i feel like we don't really gain much from doing this.

schema/schema.cql
```diff
@@ -24,3 +24,9 @@ create table metric_name_set (
   metric_names set<varchar>,
   primary key (shard)
 );
+
+-- graphite_names
+create table graphite_names (
```
jeeyoungk: i think we should do something similar as the metric_names. querying this would be quite inefficient, and I rather have some shard key to shard data around.

Nathan-Fenner: Will sharding be useful here? The only operation that we can perform is getting *all* of them, since the Go code is handling the filtering. What would the benefit of splitting them up be?

internal/cassandra.go
Nathan-Fenner: It's not really necessary (we can get rid of the hierarchy by dropping the embed and renaming it to `GraphiteStore`).

internal/api.go
```diff
@@ -110,6 +110,9 @@ func (a *defaultAPI) RemoveMetric(metric api.TaggedMetric) error {
 }
 
 func (a *defaultAPI) ToGraphiteName(metric api.TaggedMetric) (api.GraphiteMetric, error) {
+	if metric.MetricKey == "#graphite" {
```
jeeyoungk: this should be a constant and documented somewhere. also, let's use `$` as special tags. (i've done it few times in the front-end for the UI styling issues.

schema/schema.cql
jeeyoungk: it becomes easier later when we say, fetch data partially. it becomes much easier to work with partial data if it is explicitly sharded (I abused that fact several times while working with Blueflood).

api/api.go
```diff
@@ -47,6 +47,14 @@ type API interface {
 	GetMetricsForTag(tagKey, tagValue string) ([]MetricKey, error)
 }
 
+type GraphiteStore interface {
```
jeeyoungk: let's just make this into the `api/API` and figure out splitting it up later. I think we have very little things to gain by doing it (right now).

function/graphite/graphite.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/square/metrics/api"
+)
+
+const SpecialGraphiteName = "$graphite"
+
+var segmentRegex = regexp.MustCompile(`^[^.]*`)
+
+func applyPattern(pieces []string, metric string) (api.TaggedMetric, bool) {
+	tagset := api.NewTagSet()
```
jeeyoungk: is it possible to reuse `internal.Compile()` here somehow?

function/registry/registry.go
```diff
@@ -333,3 +337,62 @@ func NewOperator(op string, operator func(float64, float64) float64) function.Me
...
+		}
+		graphiteAPI, ok := context.API.(api.GraphiteStore)
+		if !ok {
+			return nil, fmt.Errorf("your API does not support graphite queries")
+		}
+		results := graphite.GetGraphiteMetrics(graphitePattern, graphiteAPI)
+		if len(results) == 0 {
+			return nil, fmt.Errorf("graphite query %s resulted in no graphite metrics", graphiteQuery)
+		}
+		ok = context.FetchLimit.Consume(len(results))
```
jeeyoungk: we should apply the predicates here, before we consume, and before we fetch. it's wasteful to fetch everything if we're going to throw them away later.

function/registry/registry.go
```diff
@@ -333,3 +337,62 @@ func NewOperator(op string, operator func(float64, float64) float64) function.Me
...
+			},
+		)
+		if err != nil {
+			return nil, err
+		}
+		for _, series := range serieslist.Series {
+			// remove the intermediate "$graphite" tag from each series
+			delete(series.TagSet, graphite.SpecialGraphiteName)
+		}
+		if context.Predicate != nil {
```
jeeyoungk: let's move this logic to top, applying against the tagsets.

function/graphite/graphite.go
```diff
@@ -0,0 +1,52 @@
...
+		Pattern:          pattern,
+		MetricKeyPattern: api.SpecialGraphiteName,
+	})
+	if err != nil {
+		return nil
+	}
+	metrics, err := API.GetAllGraphiteMetrics()
+	if err != nil {
+		// There was some issue with data
+		return nil
```
jeeyoungk: this seems a bit dangerous to silently consume the error.

function/graphite/graphite.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	results := []api.TaggedMetric{}
+	for _, metric := range metrics {
+		result, ok := applyPattern(rule, string(metric))
+		if ok {
+			results = append(results, result)
+		}
+	}
+	return results
```
jeeyoungk: i wonder how performant this is going to be - with 100k unmatched items, it can potentially take ~minutes.

we can push it to staging / prod first, but i strongly feel that we would need 1. caching and 2. parallelization.

function/graphite/graphite_test.go
```diff
@@ -0,0 +1,300 @@
...
+			t.Errorf("Test i = %d, test = %+v: didn't expect ok = %t", i, test, ok)
+			continue
+		}
+		if !test.expect.Equals(result.TagSet) {
+			t.Errorf("Expected %+v but got %+v", test.expect, result)
+		}
+	}
+}
+
+type testStore struct {
```
jeeyoungk: if you just embed the interface into it, you wouldn't need to implement (most) of the methods below.

internal/cassandra.go
jeeyoungk: what i meant was - couldn't we embed in the `type Database`?

query/command_graphite_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+import (
+	"testing"
+
+	"github.com/square/metrics/api"
+	"github.com/square/metrics/api/backend"
+	"github.com/square/metrics/assert"
+)
+
+type fakeGraphiteAPI struct {
```
jeeyoungk: i think you can do the same thing here - embed the interface, and don't implement some of the methods that are never called in the tests.

jeeyoungk:
LGTM

internal/cassandra.go
Nathan-Fenner: The Database isn't exposed in the EvaluationContext. I think we should look at eventually refactoring this, at the very least.

internal/api.go
```diff
@@ -123,5 +126,18 @@ func (a *defaultAPI) ToTaggedName(metric api.GraphiteMetric) (api.TaggedMetric,
 	return api.TaggedMetric{}, newNoMatch()
 }
 
+func (a *defaultAPI) AddGraphiteMetric(metric api.GraphiteMetric) error {
+	if graphiteDB, ok := a.db.(DatabaseGraphiteStore); ok {
+		return graphiteDB.AddGraphiteMetric(metric)
+	}
+	return nil
```
achow: Shouldn't this be an error?

internal/api.go
```diff
@@ -123,5 +126,18 @@ func (a *defaultAPI) ToTaggedName(metric api.GraphiteMetric) (api.TaggedMetric,
...
+}
+func (a *defaultAPI) GetAllGraphiteMetrics() ([]api.GraphiteMetric, error) {
+	if graphiteDB, ok := a.db.(DatabaseGraphiteStore); ok {
+		return graphiteDB.GetAllGraphiteMetrics()
+	}
+	return nil, nil
```
achow: Error?

internal/cassandra.go
```diff
@@ -197,3 +204,25 @@ func (db *defaultDatabase) RemoveFromTagIndex(tagKey string, tagValue string, me
 		tagValue,
 	).Exec()
 }
+
+func (db *defaultDatabase) AddGraphiteMetric(metric api.GraphiteMetric) error {
+	return db.session.Query(
+		"INSERT INTO graphite_names (shard, graphite_name) VALUES (?, ?) USING TTL 90000", // Slightly more than 24 hours TTL
```
achow: TTL should be configurable

internal/cassandra.go
```diff
@@ -197,3 +204,25 @@ func (db *defaultDatabase) RemoveFromTagIndex(tagKey string, tagValue string, me
...
+		0,
```
achow: There's support for sharding, but we don't really shard?

query/command_graphite_test.go
```diff
@@ -0,0 +1,350 @@
...
+			t.Errorf("Unexpected error while execution: %s", err.Error())
+			continue
+		}
+		seriesListList, ok := rawResult.([]api.SeriesList)
+		if !ok || len(seriesListList) != 1 {
+			t.Errorf("expected query `%s` to produce []value; got %+v :: %T", test.query, rawResult, rawResult)
+			continue
+		}
+		list := seriesListList[0]
+		if err != nil {
```
achow: What's this for?

query/command_graphite_test.go
```diff
@@ -0,0 +1,350 @@
...
+			t.Fatal(err)
+		}
+		a := assert.New(t)
+		expectedSeries := test.expected
```
achow: Should check the number of series you get back is the same as the number of series you expect.

query/command_graphite_test.go
Nathan-Fenner: Mistakenly left in from when another function was called; now removed in 37444f2

query/command_graphite_test.go
Nathan-Fenner: Checking for it now

internal/cassandra.go
Nathan-Fenner: There's not much benefit to sharding right now- currently, *all* graphite metrics are always fetched. I'm not sure how exactly we could avoid fetching all of them.

internal/cassandra.go
Nathan-Fenner: It's now configurable, through the API config YAML as `database: graphite_metric_ttl: (number)`.

PR #177, opened 2015-07-22
Add caching to tag conversion

Nathan-Fenner:
Whenever a metric name is received, if it succeeds at a match, its value is cached.

TODO: check if this consumes too much memory?

### DIFF
--- a/internal/api.go
+++ b/internal/api.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/gocql/gocql"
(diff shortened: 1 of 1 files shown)


internal/api.go
```diff
@@ -114,8 +119,17 @@ func (a *defaultAPI) ToGraphiteName(metric api.TaggedMetric) (api.GraphiteMetric
...
+	a.mutex.Lock()
+	if result, ok := a.cached[metric]; ok {
+		a.mutex.Unlock()
+		return result, nil
+	}
+	a.mutex.Unlock()
 	match, matched := a.ruleset.MatchRule(string(metric))
 	if matched {
+		a.mutex.Lock()
+		a.cached[metric] = match
```
jeeyoungk: failure should be cached also - otherwise failed run will continuously burn CPU.

jeeyoungk:
so... decline?

PR #179, opened 2015-07-23
ensure that shift-moving doesn't pop up the autocomplete

jeeyoungk:


Nathan-Fenner:
LGTM, but I'd like to take this as an opportunity to add "group" and "collapse" as keywords

PR #180, opened 2015-07-27
Improvement to Autocomplete Performance

Nathan-Fenner:
When there are a lot of alternatives, the autcomplete performance starts to suffer. Adding a cache should help this a little bit.

Some related changes were made to how shift behavior works (now shift+arrows will suppress the tooltip, but not shift+letter which made it impossible to autocomplete things with capital letters).

### DIFF
--- a/main/static/autocom.js
+++ b/main/static/autocom.js
@@ -304,14 +304,14 @@ function Autocom(input) {
 				tooltipSuppress = true;
 				refresh();
 				return;
-			} else if (e.keyCode == 38) { // UP
+			} else if (e.keyCode == 38 && !e.shiftKey) { // UP
 				e.preventDefault();
 				tooltipState.index--;
 				if (tooltipState.index < 0) {
 					tooltipState.index = tooltipState.words.length-1;
 				}
 				return;
-			} else if (e.keyCode == 40) { // DOWN
+			} else if (e.keyCode == 40 & &!e.shiftKey) { // DOWN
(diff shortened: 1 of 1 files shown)


main/static/autocom.js
```diff
@@ -267,9 +284,13 @@ function generateTooltipContents(tooltip, words, index, selectedCallback) {
...
 
 // Inserts `word` at the location specified by `at` inside of `input`.
-function insertWord(input, at, word) {
+function insertWord(input, at, word, supressCallback) {
 	input.value = input.value.substring(0, at.from) + word + input.value.substring(at.to);
-	input.selectionStart = input.selectionEnd = at.from + word.length;
+	setTimeout(function() {
+		input.focus();
+		input.selectionStart = input.selectionEnd = at.from + word.length;
+		supressCallback();
```
jeeyoungk: what exactly does this do?

Nathan-Fenner: I found that when you clicked on a autocomplete choice (instead of using tab/enter to fill it in), then the input box would unselected.

The call to `.focus()` brings the input back into focus. Assigning the `.selectionStart` causes the cursor to be at the end of the inserted word (instead of wherever it was before or at the end of the input).

The `suppressCallback()` has the effect of suppressing the tooltip until further input- like when you press enter to autocomplete, the suggestions go away (until you type another character).

jeeyoungk:
LGTM

PR #181, opened 2015-07-27
Change layout for describe

Nathan-Fenner:
When there were lots of tags, the columns would layout in a way that was very difficult to read and understand.

This change places the columns from a `describe ...` query into a box of fixed maximum size, showing scrollbars when its contents can't all be shown at once.

### DIFF
--- a/main/static/index.html
+++ b/main/static/index.html
@@ -91,11 +91,12 @@
         <table ng-show="isTabular() && queryResult.name != 'describe'" class="table table-striped">
           <tr ng-repeat="row in queryResult.body"><td>{{row}}</td></tr>
         </table>
-        <div ng-show="isTabular() && queryResult.name == 'describe'">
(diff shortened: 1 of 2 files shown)


achow:
Why add scrollbars?

Nathan-Fenner:
For some metrics (which have a lot of keys), the columns for the keys overflow the width of the page.

The alternatives are wrapping around (which doesn't work well because the columns can have hundreds of members), or expanding the size of the page (which just introduces a new scrollbar elsewhere).

For vertical scrollbars, I found it was helpful to be able to see the query box at all times. In particular, you don't have to bounce up and down between the keys/values you're looking at and the query box.

Nathan-Fenner:
The scrollbars are styled to be visible for the reason that Macs hide them unless you're actively scrolling, which makes it hard to notice when there's more columns or below rows below.

jeeyoungk:
LGTM, this looks very nice.

PR #182, opened 2015-07-27
Add warning on embed UI when series limit is reached

Nathan-Fenner:
Only 200 series are ever shown, even if more are returned. This is true on both the main UI view and the embedded view.

However, when the limit is enforced, the main UI view got a message, while the embedded view did not.

This PR adds a notice to embedded views when there are more than 200 series (but only the 200 are shown).

Before:
screen shot 2015-07-28 at 1 27 32 pm
After:
screen shot 2015-07-28 at 1 28 18 pm


achow:
LGTM, would be awesome if you could attach before/after screenshots for UI changes in future PRs

PR #185, opened 2015-07-28
Dynamic resolution adjustment

Nathan-Fenner:
If the given query in time requires more than 3000 points, the resolution will be dynamically lowered (they're tried in the order `user-choice`, `30s`, `5m`, `1hr`, `1d`).

3000 points at 30s: 1 day
3000 points at 5m: 10d
3000 points at 1hr: 120 days

TODO: fully integration with blueflood; a `blueflood` interface method `DecideTimerange` is used to calculate the timerange with its resolution. However, it uses hard-coded magic numbers to describe the available options instead of the defined blueflood resolutions.

TODO: configure the slot limit (currently hard-coded as 3000)

### DIFF
--- a/query/command_test.go
+++ b/query/command_test.go
@@ -435,7 +435,6 @@ func TestCommand_Select(t *testing.T) {
 				},
 			},
 		}},
-		{"select series_1 from -1000d to now resolution 30s", true, api.SeriesList{}},
 	} {
 		a := assert.New(t).Contextf("query=%s", test.query)
 		expected := test.expected
--- a/query/command_test.go
+++ b/query/command_test.go
@@ -31,6 +31,10 @@ var emptyGraphiteName = api.GraphiteMetric("")
 
 type fakeApiBackend struct{}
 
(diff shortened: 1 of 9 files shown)


Nathan-Fenner:
I had originally tried to make this change in the Blueflood fetch, but I realized that there were several problems with this approach. Namely, that the resolution of the result is fixed (so yes, I could reduce the resolution from 30s to 5min, but the user would still receive the 10,000+ points of data they (probably accidentally) asked for)

As a result, I went back to doing it at the beginning of the query, before executing anything at all.

I think that these functions could be placed inside of an interface function for `Backend` or `API`, to avoid duplication of logic/data.

query/command_test.go
```diff
@@ -435,7 +439,6 @@ func TestCommand_Select(t *testing.T) {
 				},
 			},
 		}},
-		{"select series_1 from -1000d to now resolution 30s", true, api.SeriesList{}},
```
Nathan-Fenner: This testcase causes a panic with the testing backend, so it has been removed

PR #186, opened 2015-07-29
Show error when 0 series are returned

Nathan-Fenner:
Previously there was no good indicator that no series were returned. Now there is.

An error will show for both the UI view and the embed view.

If you do a query with multiple expressions (ie `select X, Y from -10m to now`) then if either `X` or `Y` result in 0 expressions, the error will be shown. 

@jeeyoungk @achow

### DIFF
--- a/main/static/embed.html
+++ b/main/static/embed.html
@@ -17,7 +17,8 @@
     <div class="alert alert-info"   ng-show="screenState() == 'rendering'">Rendering chart...</div>
     <div class="alert alert-danger" ng-show="screenState() == 'error'">{{(queryResult.message || "").trim()}}</div>
   </div>
-  <google-chart ng-show="screenState() != 'loading' && screenState() != 'error' && queryResult.name === 'select'"
+  <div class="alert alert-danger" ng-show="screenState() == 'rendered' && queryResultIsEmpty()">Query resulted in 0 series</div>
+  <google-chart ng-show="screenState() != 'loading' && screenState() != 'error' && queryResult.name === 'select' && !queryResultIsEmpty()"
 		class="metric-chart"
 		data="selectResult"
 		option="selectOptions"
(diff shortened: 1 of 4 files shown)


main/static/script.js
```diff
@@ -301,6 +318,7 @@ module.controller("commonCtrl", function(
     if ($scope.selectResult) {
       for (var i = 0; i < queryResult.body.length; i++) {
         // Each of these is a list of series
+        console.log("?", queryResult.body[i].series.length);
```
achow: Leftover?

main/static/embed.html
```diff
@@ -17,7 +17,8 @@
     <div class="alert alert-info"   ng-show="screenState() == 'rendering'">Rendering chart...</div>
     <div class="alert alert-danger" ng-show="screenState() == 'error'">{{(queryResult.message || "").trim()}}</div>
   </div>
-  <google-chart ng-show="screenState() != 'loading' && screenState() != 'error' && queryResult.name === 'select'"
+  <div class="alert alert-danger" ng-show="screenState() == 'rendered' && queryResultIsEmpty()">{{ queryEmptyMessage }}</div>
+  <google-chart ng-show="screenState() != 'loading' && screenState() != 'error' && queryResult.name === 'select' && !queryResultIsEmpty()"
```
achow: Should indent this a bit

achow:
Is there a way to make this fit as the graph would (filling to the frame with text centered)?

Nathan-Fenner:
Yes, that would be a simple change to the `.alert-danger` style in `embed_style.css`. Right now it's 100% horizontally, but centered vertically with a fixed height (instead of 100% vertically).

achow:
Let's make it 100% vertical (or however we style the graph size).

achow:
LGTM

PR #187, opened 2015-07-29
Configure display timezones

Nathan-Fenner:
Adds a search parameter `timezone` to control the display timezone. These timezones, among many others, are supported:

* PST
* PDT
* MST
* MDT
* EST 
* EDT
* JST
* UTC

The following "magic" timezones are also supported:
* PT
* MT
* ET

When the display is specified to be one of these, either `_DT` or `_ST` corresponding to the timezone will be selected based on the user's current time of year and locale.

If you want to use a particular UTC offset, just set `location=-7` or `location=14` (to set -7 and +14 offsets respectively). These can also be fractional- `location=7.5` sets a timezone offset of `UTC+07:30`.

In addition, the special timezone `LOCAL` corresponds to local time (it's the default).

Zones are case-insensitive. 

Full support of timezone acronyms is not possible- due to the large number of worldwide timezones there are several with conflicting names. In order to support North American timezones, zones with conflicting names may be omitted.

TODO: make the timezone labels prettier

### DIFF
--- a/main/static/embed.html
+++ b/main/static/embed.html
@@ -17,6 +17,7 @@
     <div class="alert alert-info"   ng-show="screenState() == 'rendering'">Rendering chart...</div>
     <div class="alert alert-danger" ng-show="screenState() == 'error'">{{(queryResult.message || "").trim()}}</div>
   </div>
+  <div>{{ displayTimezone }}</div>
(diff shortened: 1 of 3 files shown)


syamp:
I think we should probably just use moment.js ?

PR #191, opened 2015-08-04
Fill in lower-resolution data with full resolution data

Nathan-Fenner:
Full resolution data (roughly 30s resolution) is usually available immediately. However, it takes a while for rollups to be performed and therefore lower resolution data (5m, 1hr, 1day) may not be available for a while.

This PR adds a feature to fetch full-resolution data to fill this gap in data. The last 4hrs (configurable) of the requested timerange is also fetched from full resolution data. They're combined when bucketing.

Since the full-resolution data should be sampled down in the same way that rollups are performed, it will look like the regular data is fully available. 

This essentially doubles the number of fetches required, although the number of points returned is not likely doubled.

### DIFF
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -126,7 +126,8 @@ type sampler struct {
 }
 
 // The amount of time before other resolutions become available
-const availableOnlyFull = time.Hour * 4
+// It's not "const" so that it can be mocked out for tests
+var availableOnlyFull = time.Hour * 4
 
 func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Timeseries, error) {
 	sampler, ok := samplerMap[request.SampleMethod]
@@ -154,7 +155,7 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
 
 	// The "now" as an epoch in milliseconds
 	now := time.Now().Unix() * 1000
-	earliestTime := now - int64(AvailableOnlyFull/time.Millisecond)
+	earliestTime := now - int64(availableOnlyFull/time.Millisecond)
 	latestTime := now
 
 	fullResolutionStart := fullResolutionRequest.Timerange.Start()
@@ -237,7 +238,7 @@ func (b *blueflood) fetch(request api.FetchSeriesRequest, queryUrl *url.URL) (qu
 	go func() {
 		resp, err := b.client.Get(queryUrl.String())
 		if err != nil {
-			failure <- api.BackendError{request.Metric, api.FetchIOError, "error while fetching - http connection"}
+			failure <- api.BackendError{request.Metric, api.FetchIOError, fmt.Sprintf("error while fetching - http connection [%s] error: %s", queryUrl.String(), err.Error())}
 			return
 		}
 		defer resp.Body.Close()
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -160,19 +160,19 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
 			// Clip the timerange
 			newTimerange, err := api.NewSnappedTimerange(request.Timerange.End()-b.config.FullResolutionOverlap*1000, request.Timerange.End(), request.Timerange.ResolutionMillis())
 			if err != nil {
-				log.Debugf("FULL resolution data errored while building timerange: %s", err.Error())
+				log.Infof("FULL resolution data errored while building timerange: %s", err.Error())
 				return nil
 			}
 			fullResolutionRequest.Timerange = newTimerange
 		}
 		fullResolutionQueryURL, err := b.constructURL(fullResolutionRequest, sampler, ResolutionFull)
 		if err != nil {
-			log.Debugf("FULL resolution data errored while building url: %s", err.Error())
+			log.Infof("FULL resolution data errored while building url: %s", err.Error())
 			return nil
 		}
 		fullResolutionParsedResult, err := b.fetch(request, fullResolutionQueryURL)
 		if err != nil {
-			log.Debugf("FULL resolution data errored while parsing result: %s", err.Error())
+			log.Infof("FULL resolution data errored while parsing result: %s", err.Error())
 			return nil
 		}
 		// The higher-resolution data will likely overlap with the requested data.
--- a/api/backend/blueflood/blueflood.go
+++ b/api/backend/blueflood/blueflood.go
@@ -35,10 +35,11 @@ type httpClient interface {
 }
 
 type Config struct {
-	BaseUrl  string           `yaml:"base_url"`
-	TenantId string           `yaml:"tenant_id"`
-	Ttls     map[string]int64 `yaml:"ttls"` // Ttl in days
-	Timeout  time.Duration    `yaml:"timeout"`
+	BaseUrl               string           `yaml:"base_url"`
+	TenantId              string           `yaml:"tenant_id"`
+	Ttls                  map[string]int64 `yaml:"ttls"` // Ttl in days
+	Timeout               time.Duration    `yaml:"timeout"`
+	FullResolutionOverlap int64            `yaml:"full_resolution_overlap"` // overlap to draw full resolution in seconds
 }
 
 func (c Config) getTTL(r Resolution) time.Duration {
@@ -125,10 +126,6 @@ type sampler struct {
 	bucketSampler func([]float64) float64
 }
 
-// The amount of time before other resolutions become available
-// It's not "const" so that it can be mocked out for tests
-var availableOnlyFull = time.Hour * 4
-
 func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Timeseries, error) {
 	sampler, ok := samplerMap[request.SampleMethod]
 	if !ok {
@@ -152,11 +149,24 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
 	combinedResult := parsedResult.Values
 
 	// Sample the data at the FULL resolution.
-	// In order to do this, we use the same `request` object but hard-code the ResolutionFull parameter.
-	fullResolutionQueryURL, err := b.constructURL(request, sampler, ResolutionFull)
+	// We clip the timerange so that it's only #{config.FullResolutionOverlap} seconds long.
+	// This limits the amount of data to be fetched.
+
+	fullResolutionRequest := request
+	if (request.Timerange.End()-request.Timerange.Start())/1000 > b.config.FullResolutionOverlap {
+		// Clip the timerange
+		newTimerange, err := api.NewSnappedTimerange(request.Timerange.End()-b.config.FullResolutionOverlap*1000, request.Timerange.End(), request.Timerange.ResolutionMillis())
+		if err == nil {
+			fullResolutionRequest.Timerange = newTimerange
+		}
+	}
+	fullResolutionQueryURL, err := b.constructURL(fullResolutionRequest, sampler, ResolutionFull)
 	if err == nil {
 		fullResolutionParsedResult, err := b.fetch(request, fullResolutionQueryURL)
 		if err == nil {
+			// The higher-resolution data will likely overlap with the requested data.
+			// This isn't a problem - the requested, higher-resolution data will be downsampled by this code.
+			// This downsampling should arrive at the same answer as Blueflood's built-in rollups.
 			combinedResult = append(parsedResult.Values, fullResolutionParsedResult.Values...)
 		}
 	}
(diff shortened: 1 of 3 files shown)


api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +135,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+	}
+	if fullResolutionEnd > latestTime {
+		fullResolutionEnd = latestTime
+	}
 
-	values := processResult(parsedResult, request.Timerange, sampler, queryResolution)
+	var fullResolutionParsedResult queryResponse
+	fullResolutionRequest.Timerange, err = api.NewSnappedTimerange(fullResolutionStart, fullResolutionEnd, 30000)
+	if err != nil {
+		fmt.Printf("<<error snapping timerange for 30s>>\n")
```
jeeyoungk: can we log the error instead of `fmt.Printf?`

jeeyoungk:
i think we should be able to test behavior by seeding the `fakeHttpClient` with more data.

api/backend/blueflood/blueflood.go
```diff
@@ -125,6 +125,9 @@ type sampler struct {
 	bucketSampler func([]float64) float64
 }
 
+// The amount of time before other resolutions become available
+var AvailableOnlyFull = time.Hour * 4
```
jeeyoungk: lowercase & const?

Nathan-Fenner:
Tests are implemented by very carefully using `time.Now()` to make fake data that's set about 30m in the past.

There is potential for these tests to be flaky; on my system, running this set of tests takes about 0.1 seconds. If evaluation is split over two different 30s intervals (for instance, starting at 3:47:29.95 and ending at 3:47:30.05) then the tests may fail.

In an actual deployment, the data may be slightly shifted, but no failure would occur.

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
 	}
 	queryResolution := b.config.bluefloodResolution(
 		request.Timerange.Resolution(),
-		request.Timerange.Start())
+		request.Timerange.Start(),
+	) // The query resolution will be downsampled automatically based on the timerange?
```
achow: What does this comment mean?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
 	if err != nil {
 		return api.Timeseries{}, err
 	}
-
-	// Issue GET to fetch metrics
 	parsedResult, err := b.fetch(request, queryUrl)
 	if err != nil {
 		return api.Timeseries{}, err
 	}
+	// </Sample the data>
```
achow: Probably don't need a comment here about the end of a logical block. You can add a comment at the beginning and it should be fine.

jeeyoungk:
LGTM

api/backend/blueflood/blueflood.go
```diff
@@ -125,6 +129,10 @@ type sampler struct {
 	bucketSampler func([]float64) float64
 }
 
+// The amount of time before other resolutions become available
+// It's not "const" so that it can be mocked out for tests
+var availableOnlyFull = time.Hour * 4
```
achow: Is this specific to our setup, or does blueflood statically set the thresholds for when lower resolution data is available?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
 	if err != nil {
 		return api.Timeseries{}, err
 	}
+	// </Sample the data>
+
+	// <Sample the full resolution data>
+	fullResolutionRequest := request
+
+	// The "now" as an epoch in milliseconds
+	now := getNow().Unix() * 1000
```
achow: Why do we need the current timestamp? Shouldn't we be able to infer the beginning and end range of full resolution data we need to fetch?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+	fullResolutionEnd := fullResolutionRequest.Timerange.End()
+	if fullResolutionEnd < earliestTime {
+		fullResolutionEnd = earliestTime
+	}
+	if fullResolutionEnd > latestTime {
+		fullResolutionEnd = latestTime
+	}
+
+	var fullResolutionParsedResult queryResponse
+	fullResolutionRequest.Timerange, err = api.NewSnappedTimerange(fullResolutionStart, fullResolutionEnd, 30000)
```
achow: Why are we hardcoding a 30 second resolution here?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+	if err != nil {
+		log.Errorf("Error building full-resolution timerange: %s", err.Error())
```
achow: Shouldn't return an error at this point?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+	} else {
+		fullResolutionQueryUrl, err := b.constructURL(fullResolutionRequest, sampler, ResolutionFull)
+		if err != nil {
+			log.Errorf("Error building full-resolution URL: %s", err.Error())
```
achow: Error?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +140,61 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+		} else {
+			fullResolutionParsedResult, err = b.fetch(fullResolutionRequest, fullResolutionQueryUrl)
+			if err != nil {
+				log.Errorf("Error parsing full-resolution query response: %s", err.Error())
```
achow: Error?

api/backend/blueflood/blueflood.go
Nathan-Fenner: The threshold I used is arbitrary. I don't know when Blueflood guarantees rollups will have occurred. A much smaller number (1hr?) in practice would still probably work.

api/backend/blueflood/blueflood.go
```diff
@@ -223,13 +274,21 @@ func (b *blueflood) fetch(request api.FetchSeriesRequest, queryUrl *url.URL) (qu
 	}
 }
 
-func processResult(parsedResult queryResponse,
+func processResult(
+	parsedResult queryResponse,
+	fullResolutionParsedResult queryResponse,
```
achow: Can we merge these together before the call to processResult()?

api/backend/blueflood/blueflood.go
```diff
@@ -271,10 +330,14 @@ var samplerMap map[api.SampleMethod]sampler = map[api.SampleMethod]sampler{
 		fieldSelector: func(point metricPoint) float64 { return point.Average },
 		bucketSampler: func(bucket []float64) float64 {
 			value := 0.0
+			count := 0.0
```
achow: Let's make this an integer and coerce to float at divide time

api/backend/blueflood/blueflood.go
Nathan-Fenner: The goal of clipping the data to be recent-ish was to reduce the amount of unneeded full-resolution points being requested. 

It's probably simpler to just request the same full timerange for full resolution data.

Will there be any performance impact from this? Or will this not be a problem?

api/backend/blueflood/blueflood.go
```diff
@@ -283,6 +346,12 @@ var samplerMap map[api.SampleMethod]sampler = map[api.SampleMethod]sampler{
 		bucketSampler: func(bucket []float64) float64 {
 			value := bucket[0]
 			for _, v := range bucket {
+				if math.IsNaN(v) {
+					continue
+				}
+				if math.IsNaN(value) {
+					value = v
+				}
 				value = math.Min(value, v)
```
achow: Should probably put this line an an else. It ends up being a noop when `IsNaN(value)`

api/backend/blueflood/blueflood.go
```diff
@@ -294,6 +363,12 @@ var samplerMap map[api.SampleMethod]sampler = map[api.SampleMethod]sampler{
 		bucketSampler: func(bucket []float64) float64 {
 			value := bucket[0]
 			for _, v := range bucket {
+				if math.IsNaN(v) {
+					continue
+				}
+				if math.IsNaN(value) {
+					value = v
+				}
 				value = math.Max(value, v)
```
achow: Same as above.

api/backend/blueflood/blueflood.go
achow: If it's arbitrary, let's make this a `blueflood` field so it's configurable.

api/backend/blueflood/blueflood_test.go
```diff
@@ -252,3 +254,276 @@ func TestSeriesFromMetricPoints(t *testing.T) {
...
+		"https://blueflood.url",
+		"square",
+		make(map[string]int64),
+		time.Millisecond,
+	}
+
+	rightNow := time.Now()
+	// Stop time from moving forward by replacing `getNow` with this:
+	// (which prevents hard-to-find timing bugs that would cause the test to be flaky)
+	getNow = func() time.Time {
```
achow: Oh wow, that's quite the abuse of package-private globals.

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +136,32 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
 		return api.Timeseries{}, err
 	}
 
-	values := processResult(parsedResult, request.Timerange, sampler, queryResolution)
+	combinedResult := parsedResult.Values
+
+	// Sample the data at the FULL resolution.
+	// In order to do this, we use the same `request` object but hard-code the ResolutionFull parameter.
+	fullResolutionQueryURL, err := b.constructURL(request, sampler, ResolutionFull)
+	if err == nil {
```
achow: Shouldn't failure error? Or is there a reason we shouldn't care?

api/backend/blueflood/blueflood.go
```diff
@@ -132,20 +136,32 @@ func (b *blueflood) FetchSingleSeries(request api.FetchSeriesRequest) (api.Times
...
+		fullResolutionParsedResult, err := b.fetch(request, fullResolutionQueryURL)
+		if err == nil {
+			combinedResult = append(parsedResult.Values, fullResolutionParsedResult.Values...)
```
achow: Should add a comment here about how overlapping data will end up being the same as blueflood's rollups after we do our own. It's not completely intuitive that that's the case.

There may be some situations that this ends up giving wrong data, but I can't think of any. I'm relatively confident this works out though.

api/backend/blueflood/blueflood.go
Nathan-Fenner: I'm not sure.

On the one hand, I don't expect this part to error. On the other hand, if it does, the query isn't really broken- if you're querying for low-resolution data it's a bit strange if the high-resolution data breaks it.

Also, from a practical effort standpoint, the mocking tests for Blueflood fail since they don't have a `constructURL` for the corresponding high-resolution fetch.

Nathan-Fenner:
`peg` has updated.

api/backend/blueflood/blueflood_test.go
```diff
@@ -252,3 +254,150 @@ func TestSeriesFromMetricPoints(t *testing.T) {
...
+		baseTime-300*1000*8,  // 40 minutes ago
+		baseTime-300*1000*7,  // 35 minutes ago
+	)
+
+	fullResolutionQueryURL := fmt.Sprintf(
+		"https://blueflood.url/v2.0/square/views/some.key.value?from=%d&resolution=FULL&select=numPoints%%2Caverage&to=%d",
+		baseTime-300*1000*10, // 50 minutes ago
+		baseTime-300*1000*3,  // 15 minutes ago
+	)
+	fmt.Printf("expect full [%s]\n", fullResolutionQueryURL)
```
achow: Remove?

api/backend/blueflood/blueflood_test.go
```diff
@@ -252,3 +254,150 @@ func TestSeriesFromMetricPoints(t *testing.T) {
...
+	}
+
+	baseTime := 1438734300000
+
+	regularQueryURL := fmt.Sprintf(
+		"https://blueflood.url/v2.0/square/views/some.key.value?from=%d&resolution=MIN5&select=numPoints%%2Caverage&to=%d",
+		baseTime-300*1000*10, // 50 minutes ago
+		baseTime-300*1000*3,  // 15 minutes ago
+	)
+	fmt.Printf("expect regular [%s]\n", regularQueryURL)
```
achow: Remove?

achow:
LGTM after removing some debug in tests.

PR #193, opened 2015-08-25
Expose the full query even if an alias was used

drcapulet:
This introduces a new field on `SeriesList` named `Query` that has the full query for the given TimeSeries. This will be the same as `Name` unless `transform.alias` is used.


syamp:
lgtm - curios on test failures tho

drcapulet:
Yeah, it's failing on master as well - looks like this commit: https://github.com/pointlander/peg/commit/cafbe8e1a27fbf0b80d4ccfcce051b6714151205

drcapulet:
Fixed in #194 and rebased.

PR #196, opened 2015-08-27
Add bulk add Metrics API and remove db-specific caching.

cchandler:
This adds a simpler bulk add metrics API for cassandra that does streaming. Doesn't look like the go cassandra driver supports the same level of streaming large inserts that the Java/C# driver does (https://medium.com/@foundev/cassandra-batch-loading-without-the-batch-keyword-40f00e35e23e). I could be wrong. Either way this is about as fast as I can get it without relying on more clever methods.

I'm also removing the caching from the cassandra code in favor of moving it higher up into the stack.

### DIFF
--- a/internal/cassandra.go
+++ b/internal/cassandra.go
@@ -16,8 +16,6 @@
 package internal
 
 import (
-	"sync"
-
 	"github.com/gocql/gocql"
 	"github.com/square/metrics/api"
 )
@@ -27,6 +25,7 @@ type Database interface {
 	// Insertion Methods
 	// -----------------
 	AddMetricName(metricKey api.MetricKey, metric api.TagSet) error
+	AddMetricNames(metrics []api.TaggedMetric) error
 	AddToTagIndex(tagKey, tagValue string, metricKey api.MetricKey) error
 
 	// Query methods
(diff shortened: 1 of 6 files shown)


syamp:
I believe cassandra actually discourages batching as means to improve performance. @achow has more deets.

cchandler:
@syamp You should read the article I pasted ;-). It's specifically about batch loading without cassandra's batch keyword.

cchandler:
The test env is apparently flakey.

https://travis-ci.org/square/metrics/builds/77549233

internal/cassandra.go
```diff
@@ -76,44 +67,63 @@ func (db *defaultDatabase) AddMetricName(metricKey api.MetricKey, tagSet api.Tag
...
-	if indexValue {
-		return nil // Found in the cache so already in the table, so no need to perform a write.
+func (db *defaultDatabase) AddMetricNames(metrics []api.TaggedMetric) error {
+	queryInsert := "INSERT INTO metric_names (metric_key, tag_set) VALUES (?, ?)"
+	queryUpdate := "UPDATE metric_name_set SET metric_names = metric_names + ? WHERE shard = ?"
+
+	c := make(chan *gocql.Query, 10)
+	done := make(chan bool)
+	go func() {
+		for {
```
drcapulet: You can do:
```
go func() {
  for boundQuery := range c {
    _ = boundQuery.Exec()
  }
  done <- true
}()
```

drcapulet:
:+1:

PR #197, opened 2015-09-01
Don't format responses with extra spacing

drcapulet:
We tend to produce very large responses, nicely formatting the JSON wastes a ton of time. `python -mjson.tool` is your friend.


drcapulet:
Going to decline and make this a little better.

PR #199, opened 2015-09-02
Rename components and annotate

cchandler:
These are the changes @cuzelac  + @cchandler have been working on. This is entirely to make the actual larger changes coming, easier. No tests were modified except where absolutely necessary to mesh with the new APIs. A few tests were discovered to not actually cover anything and so it's possible we've introduced some problems. Starting tomorrow we're going to create a branch on our internal(stash) repos that merge these changes in. We won't merge here until those match so we can do it mostly-atomically. Talk to either me or cu if you have questions.

I'm sorry. I'm sorry. I'm sorry. I'm sorry. But it's happening.

---

NO TESTS WERE CHANGED. This is strictly a structural remap so we can start introducing changes.

Wider changes:
* Mostly renamed stuff.
* Anywhere that used a singular/sequential Backend now uses the parallel one

Specifics:
* api/api -> gone. Moved to individual pieces
* api/backend -> gone. Moved to individual pieces
* api/backend/blueflood -> timeseries_storage/blueflood
* api/backend/multibackend -> ParallelTimeseriesStorageWrapper and the sequential/singular removed.
& api/backend/multibackend -> Added guard condition if a Cancellable was not included. Induces panic.
* api/backend/multibackend -> ProfilingBackend became the ProfilingMultiTimeseriesStorageWrapper
* api/errors -> Errors grouped with their segment APIs
* api/metric_metadata_api -> New API along with ProfilingMetricMetadataAPI
* api/timeseries_storage_api -> New API
* api/timeseries_storage_api -> Old BackendErrorCode now TimeseriesStorageErrorCode
* api/types.go -> Moved to individual files for testability
* assert -> moved to testing_support/assert. No changes.
* most of function package -> Just updated tests
* internal/api.go -> Renamed to cassandra since that's all it implemented
* internal/api.go -> Moved all mentions of graphite to package util and implemented a graphite rule converter
* internal/errors.go -> Errors moved to util because they corresponded with graphite rules
* internal/rules.go -> moved to util
* main/build_mains -> Changed to correctly cd into each directory and build main
* main/* -> Updated for new names/paths
* mocks/* -> Moved to testing_support/mocks
* testing_support/mocks -> Added 'AddPairWithoutGraphite' and 'AddPair'. Now that graphite converter is its own thing, this messy hack was left as a way to also update the graphite converter along with the metadata API in testing. This will change.
* query -> Updated for new names/paths
* query/inspect_test.go -> This appears to have always been broken... Added notes about fixing it.
* test_all.sh -> Wrote a script to go through every directory with go files and run the tests
* timeseries_storage/blueflood -> Blueflood's config now takes a GraphiteConverter because the MetricMetadataAPI no longer provides it and thus cannot be reached via the FetchTimeseries* Request objects. But it's required, because we don't actually have a generic blueflood implementation- graphite names must be reconstructed at query time.
* util/graphite_converter -> New graphite converter that takes []string and produces api.TaggedMetric and in reverse. This didn't make sense as part of a generic MetricMetadataAPI and is going to be phased out into a util/contrib package. At some point we have to remove references for this from the blueflood implementation.


api/metric_metadata_api.go
```diff
@@ -16,86 +16,66 @@
...
-	Hosts    []string `yaml:"hosts"`
-	Keyspace string   `yaml:"keyspace"`
-}
-
-// ProfilingAPI wraps an ordinary API and also records profiling metrics to a given Profiler object.
-type ProfilingAPI struct {
-	Profiler *inspect.Profiler
-	API      API
+type ProfilingMetricMetadataAPI struct {
+	Profiler       *inspect.Profiler
```
drcapulet: Interesting how we have the profiler inline on the struct, whereas above in `api/backend/multibackend.go` we pull it from `request.Profiler`

api/sample_method.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package api holds common data types and public interface exposed by the indexer library.
+// Refer to the doc
+// https://docs.google.com/a/squareup.com/document/d/1k0Wgi2wnJPQoyDyReb9dyIqRrD8-v0u8hz37S282ii4/edit
```
drcapulet: (I know this is copied from elsewhere) Non-public link though

api/sample_method.go
```diff
@@ -0,0 +1,50 @@
...
+// For example, Blueflood may stores (min / max / average / count) during rollups,
+// and these data are exposed via columns
+type SeriesType string
+
+// SampleMethod determines how the given time series should be sampled.
+type SampleMethod int
+
+const (
+	// SamplingMax chooses the maximum value.
+	SampleMax SampleMethod = iota + 1
```
drcapulet: Perfect, lets not repeat the proto value 0 issues.

api/tagset.go
```diff
@@ -0,0 +1,115 @@
...
+	for key, value := range other {
+		result[key] = value
+	}
+	for key, value := range tagSet {
+		result[key] = value
+	}
+	return result
+}
+
+//TODO(cchandler): What is this? Should it be moved or deleted?
```
drcapulet: Seems like this is parsing something like `taga=5,tagb=yellow`

api/timerange.go
```diff
@@ -0,0 +1,133 @@
...
+	// Adding `boundary/2` changes this instead to a "round to nearest" rather than "round towards 0".
+	// (Where "up" is the round for values exactly halfway between).
+	// These halfway points round "away from zero" (rather than "towards -infinity").
+	return (n + boundary/2) / boundary * boundary
+}
+
+// Round() will fix some invalid timeranges by rounding their starts and ends.
+func (tr Timerange) Snap() Timerange {
+	if tr.resolution == 0 {
+		panic("AAH")
```
drcapulet: Uh?

api/tagset.go
cchandler: Yeah, the note was made a few days ago before I got that far. I'll remove my note :-)

api/timerange.go
cchandler: Yeah, that's a solid candidate for an actual error ;-)

main/metadata/metadata.go
```diff
@@ -0,0 +1,54 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	// "net/http"
```
drcapulet: nit: looks like these can be cleaned out?

drcapulet:
Any reason the `api/backend` package stuck around and wasn't moved to `timeseries_storage`?

api/backend/multibackend.go
```diff
@@ -18,52 +18,28 @@ import (
...
-}
-
-type parallelMultiBackend struct {
-	limit   int
-	tickets chan struct{}
-	Backend api.Backend
-}
-
-func NewParallelMultiBackend(backend api.Backend, limit int) api.MultiBackend {
+func NewParallelMultiBackend(backend api.TimeseriesStorageAPI, limit int) *ParallelTimeseriesStorageWrapper {
```
drcapulet: Hate the naming here of `Backend`, but we can move this later.

test_all.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
```
drcapulet: Curious, why not `go test ./...`?

drcapulet:
Overall, LGMT :+1:

test_all.sh
cchandler: It required that we fix some things everywhere at once to get useful output. Now that we've done that we definitely don't need this.

api/backend/multibackend.go
cchandler: Yeah, individual variable renames were done where it was appropriate, but sometimes it was just bogging things down too much. After this change we can go through with a grep -Ri backend and change all these.

api/sample_method.go
jeeyoungk: yes that was done on purpose.

api/sample_method.go
jeeyoungk: yes. i don't think we have many non-public facing doc (yet). maybe we can markdown & publish it later?

PR #200, opened 2015-09-08
Multirequest improvements and the end of backend

cchandler:
This will remove the backend package entirely. The API for TimeseriesStorageAPI has been expanded to include both single and multi timeseries get operations. This will move the concern of parallelism into the time series provider. In our case, this exterior wrapper works because we only rely on blueflood. Ideally this behavior is accounted for in the provider (eg a file-system based provider wanting to do something smart and coalesce the operations in a less isolated way). Unfortunately, testing that within the blueflood code is going to take some more work, but I want to start approval up to this point at least.

This also simplifies decomposing a FetchMultipleTimeseriesRequest into single requests. Now it returns an array of equivalent operations.

Tests that relied on fakeBackend are being transitioned to prefer the mock package.

Blueflood's internal getTtl was renamed to have a more obvious function. This also fixes broken tests that started failing because the ttl from a set date finally expired.


query/inspect_test.go
```diff
@@ -18,13 +18,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/square/metrics/api/backend"
+	// "github.com/square/metrics/api/backend"
```
drcapulet: nit: remove

drcapulet:
Overall :+1:

query/inspect_test.go
cchandler: I'll deal with this in the next PR. The entire thing will be around this specific file.

PR #201, opened 2015-09-09
Build txt file reports from what matches and what doesn't

cchandler:
I've modified the old ruletester file to be able to read the new zlib compressed gob encoded cache file we generate from the metrics-indexer upstream. Yes, the ruletester file is a bit of a mess right now, but I think that's fine. We need to refine this and get more data out of it most likely, so I'm leaving it there for now for the next person :-).

I've also exported a lot of the data from the util.GraphiteConverter stuff so we can get per-rule statistics from within the converter itself. There's probably more we want to track.

### DIFF
--- a/main/ruletester/ruletester.go
+++ b/main/ruletester/ruletester.go
@@ -162,10 +162,10 @@ func DoAnalysis(metrics []string, graphiteConverter util.RuleBasedGraphiteConver
 	var wg sync.WaitGroup
 
 	i := 0
-	for i = 0; i < len(metrics); i = i + 25 {
+	for i = 0; i+25 < len(metrics); i = i + 25 {
 		workQueue <- metrics[i : i+25]
 	}
-	if i <= len(metrics) {
+	if i < len(metrics) {
 		workQueue <- metrics[i:]
 	}
 
--- a/main/ruletester/ruletester.go
+++ b/main/ruletester/ruletester.go
@@ -154,8 +154,11 @@ type ChunkResult struct {
 }
 
 func DoAnalysis(metrics []string, graphiteConverter util.RuleBasedGraphiteConverter) {
+	graphiteConverter.EnableStats()
+
 	workQueue := make(chan []string, len(metrics)/25+1)
 	resultQueue := make(chan ChunkResult, len(metrics))
+	unmatchedQueue := make(chan string, len(metrics))
 	var wg sync.WaitGroup
 
 	i := 0
@@ -194,6 +197,7 @@ func DoAnalysis(metrics []string, graphiteConverter util.RuleBasedGraphiteConver
 							}
 							chunk_result.matched++
 						} else {
+							unmatchedQueue <- metric
 							chunk_result.unmatched++
 						}
 					}
@@ -211,6 +215,7 @@ func DoAnalysis(metrics []string, graphiteConverter util.RuleBasedGraphiteConver
 	}
 	wg.Wait()
 	close(resultQueue)
+	close(unmatchedQueue)
 
 	fmt.Printf("\n")
 	fmt.Printf("Processing results!")
@@ -232,6 +237,20 @@ func DoAnalysis(metrics []string, graphiteConverter util.RuleBasedGraphiteConver
 			}
 		}
 	}()
+	wg.Add(1)
+	unmatchedResults := []string{}
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case unmatched, more := <-unmatchedQueue:
+				unmatchedResults = append(unmatchedResults, unmatched)
+				if !more {
+					return
+				}
+			}
+		}
+	}()
 
 	wg.Wait()
 	fmt.Printf("\n")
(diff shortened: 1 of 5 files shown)


main/ruletester/ruletester.go
```diff
@@ -72,132 +105,295 @@ func main() {
...
+
+	workQueue := make(chan []string, len(metrics)/25+1)
+	resultQueue := make(chan ChunkResult, len(metrics))
+	unmatchedQueue := make(chan string, len(metrics))
 	var wg sync.WaitGroup
-	stat := Statistics{
-		perMetric: make(map[api.MetricKey]PerMetricStatistics),
+
+	i := 0
+	for i = 0; i < len(metrics); i = i + 25 {
```
drcapulet: This panics (example: http://play.golang.org/p/iYa3ZJCf_a).

I assume you want slice of 25 - this code would work:

```
for i = 0; i + 25 <= len(metrics); i = i + 25 {
  workQueue <- metrics[i : i+25]
}
if i < len(metrics) {
  workQueue <- metrics[i:]
}
```

util/rules.go
```diff
@@ -144,9 +175,9 @@ func (rule Rule) ToGraphiteName(taggedMetric api.TaggedMetric) (GraphiteMetric,
 
 // MatchRule sees if a given graphite string matches
 // any of the specified rules.
-func (ruleSet RuleSet) MatchRule(input string) (api.TaggedMetric, bool) {
-	for _, rule := range ruleSet.rules {
-		value, matched := rule.MatchRule(input)
+func (ruleSet *RuleSet) MatchRule(input string) (api.TaggedMetric, bool) {
+	for i := 0; i < len(ruleSet.Rules); i++ {
```
drcapulet: Why not the default `range` here?

cchandler: Because I want to modify the underlying struct on a successful match downstream and make sure it stays in memory. Range has weird copy semantics http://stackoverflow.com/questions/15945030/change-values-while-iterating-in-golang . Unless I'm missing something.

main/ruletester/ruletester.go
cchandler: Good catch. Updated.

util/rules.go
drcapulet: Ah yeah, if you want to modify the item then you do need to do it this way.

drcapulet:
:+1: LGTM

main/ruletester/ruletester.go
```diff
@@ -57,10 +60,40 @@ type PerMetricStatistics struct {
...
 }
 
+func ReadMetricsFile(file string) ([]string, error) {
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]int)
+	r, err := zlib.NewReader(bytes.NewBuffer(data))
```
Nathan-Fenner: `err` isn't being checked

main/ruletester/ruletester.go
```diff
@@ -57,10 +60,40 @@ type PerMetricStatistics struct {
...
+	io.Copy(b, r)
+	d := gob.NewDecoder(b)
+
+	err = d.Decode(&result)
+	if err != nil {
+		return nil, err
+	}
+
+	strings := []string{}
+	for k, _ := range result {
```
Nathan-Fenner: `for k := range result {` would be equivalent (and probably more idiomatic)

main/ruletester/ruletester.go
```diff
@@ -72,132 +105,295 @@ func main() {
...
+	// matched := 0
+	// unmatched := 0
+	// reverse_convert_failed := 0
+
+	// fmt.Printf("Matched: %d\n", matched)
+	// fmt.Printf("Unmatched: %d\n", unmatched)
+	// fmt.Printf("Reverse convert failed: %d\n", reverse_convert_failed)
 
-	metricFile, err := os.Open(*metricsFile)
 	if err != nil {
```
Nathan-Fenner: This error doesn't correspond to anything anymore (the previous call was deleted)

main/ruletester/ruletester.go
```diff
@@ -72,132 +105,295 @@ func main() {
...
+							}
+							chunk_result.matched++
+						} else {
+							unmatchedQueue <- metric
+							chunk_result.unmatched++
+						}
+					}
+					resultQueue <- chunk_result
+
+					if !more {
```
Nathan-Fenner: Should be at the top of this `case`. If `more` is false, then the value `metrics` will be the zero value (nil), so looping over it is pointless.

main/ruletester/ruletester.go
```diff
@@ -72,132 +105,295 @@ func main() {
...
 				select {
-				case <-done:
-					return
-				case input := <-inputBuffer:
-					metric, matched := ruleset.MatchRule(input)
-					outputBuffer <- result{
-						input,
-						metric,
-						matched,
+				case metrics, more := <-workQueue:
```
Nathan-Fenner: The `for { select {` can just be changed into a range over the channel.

`for metrics := range workQueue {`

main/ruletester/ruletester.go
Nathan-Fenner: If `more` is false, then the value extracted from the channel is the *zero value* for the type. In this case, the last value sent on the channel before it was closed has already been processed.

PR #202, opened 2015-09-09
Better error message for Timerange#Snap()

drcapulet:
R: @cchandler || @cuzelac || @jeeyoungk || @achow


drcapulet:
Declining in favor of a bigger PR around `golint`

PR #204, opened 2015-09-10
Calculate full timeranges for rates & derivatives, change Integral & Cumulative

drcapulet:
The transform.rate and transform.derivative functions now fetch an extra point
before the start of the timerange so that the start of the timerange represents
the delta from the point before, not simply 0.

The transform.integral and transform.cumulative functions now always start at
zero since the left-most datapoint isn't technically part of the series we're
summing or integrating over.

R: @cchandler || @cuzelac

### DIFF
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -174,9 +174,11 @@ func derivative(values []float64, parameters []function.Value, scale float64) ([
 	return result, nil
 }
 
-// Rate is special because it needs to get one extra data point to the left
-// This transform functions exactly like Derivative but bounds the result to be positive.
-// That is, it returns consecutive differences which are at least 0.
+// Rate is special because it needs to get one extra data point to the left.
+// This transform functions mostly like Derivative but bounds the result to be positive.
+// Specifically this function is designed for strictly increasing counters that
+// only decrease when reset to zero. That is, thie function returns consecutive
+// differences which are at least 0, or math.Max of the newly reported value and 0
 var Rate = newDerivativeBasedTransform("rate", rate)
 
 func rate(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -191,6 +191,8 @@ func rate(values []float64, parameters []function.Value, scale float64) ([]float
 		result[i-1] = (values[i] - values[i-1]) / scale
 		if result[i-1] < 0 {
 			// values[i] is our best approximatation of the delta between i-1 and i
+			// Why? This should only be used on counters, so if v[i] - v[i-1] < 0 then
+			// the counter has reset, and we know *at least* v[i] increments have happened
 			result[i-1] = math.Max(values[i], 0) / scale
 		}
 	}
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -210,6 +210,9 @@ func newDerivativeBasedTransform(name string, transformer transform) function.Me
 			newContext := context
 			timerange := context.Timerange
 			newContext.Timerange, err = api.NewSnappedTimerange(timerange.Start()-timerange.ResolutionMillis(), timerange.End(), timerange.ResolutionMillis())
+			if err != nil {
+				return nil, err
+			}
 
 			// The new context has a timerange which is extended beyond the query's.
 			listValue, err := arguments[0].Evaluate(newContext)
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -157,3 +157,90 @@ var Alias = function.MetricFunction{
 		return list, nil
 	},
 }
+
+// Derivative is special because it needs to get one extra data point to the left
+// This transform estimates the "change per second" between the two samples (scaled consecutive difference)
+var Derivative = newDerivativeBasedTransform("derivative", derivative)
+
+func derivative(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
+	result := make([]float64, len(values)-1)
+	for i := range values {
+		if i == 0 {
+			continue
+		}
+		// Scaled difference
+		result[i-1] = (values[i] - values[i-1]) / scale
+	}
+	return result, nil
+}
+
+// Rate is special because it needs to get one extra data point to the left
+// This transform functions exactly like Derivative but bounds the result to be positive.
+// That is, it returns consecutive differences which are at least 0.
+var Rate = newDerivativeBasedTransform("rate", rate)
+
+func rate(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
+	result := make([]float64, len(values)-1)
+	for i := range values {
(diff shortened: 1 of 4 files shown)


function/transform/special.go
```diff
@@ -157,3 +157,90 @@ var Alias = function.MetricFunction{
...
+	return function.MetricFunction{
+		Name:         "transform." + name,
+		MinArguments: 1,
+		MaxArguments: 1,
+		Compute: func(context function.EvaluationContext, arguments []function.Expression, groups function.Groups) (function.Value, error) {
+			var err error
+			// Calcuate the new timerange to include one extra point to the left
+			newContext := context
+			timerange := context.Timerange
+			newContext.Timerange, err = api.NewSnappedTimerange(timerange.Start()-timerange.ResolutionMillis(), timerange.End(), timerange.ResolutionMillis())
```
Nathan-Fenner: `err` isn't being checked.

An error is very unlikely, but it would occur for timeranges obscenely far in the past, such that the Start time wraps around. The resulting timerange in that case would be invalid, which could cause panics down the line due to division by zero.

If we just don't want to worry about this case, then it would be better to explicitly ignore the error with `_`, and then you don't need to define `var err error` at the top.

drcapulet: Nice catch, fixed.

function/transform/special.go
```diff
@@ -157,3 +157,93 @@ var Alias = function.MetricFunction{
...
+var Rate = newDerivativeBasedTransform("rate", rate)
+
+func rate(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
+	result := make([]float64, len(values)-1)
+	for i := range values {
+		if i == 0 {
+			continue
+		}
+		// Scaled difference
+		result[i-1] = (values[i] - values[i-1]) / scale
```
achow: Why not do a `math.Max` around the result here instead of down below?

function/transform/special.go
```diff
@@ -157,3 +157,93 @@ var Alias = function.MetricFunction{
...
+		if result[i-1] < 0 {
+			// values[i] is our best approximatation of the delta between i-1 and i
```
achow: Heh, approximatation

function/transform/special.go
```diff
@@ -157,3 +157,93 @@ var Alias = function.MetricFunction{
...
+	}
+	return result, nil
+}
+
+// Rate is special because it needs to get one extra data point to the left
+// This transform functions exactly like Derivative but bounds the result to be positive.
+// That is, it returns consecutive differences which are at least 0.
+var Rate = newDerivativeBasedTransform("rate", rate)
+
+func rate(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
```
achow: This could just call into `derivative` and `max(val, 0)` over the resulting series.

function/transform/special.go
```diff
@@ -157,3 +157,93 @@ var Alias = function.MetricFunction{
...
+
+			result, err := ApplyTransform(list, transformer, []function.Value{})
+			if err != nil {
+				return nil, err
+			}
+
+			// Validate our series are the correct length
+			for i := range result.Series {
+				if len(result.Series[i].Values) != len(list.Series[i].Values)-1 {
+					return nil, fmt.Errorf("Expected transform to return %d values, received %d", len(list.Series[i].Values)-1, len(result.Series[i].Values))
```
achow: Might be better off as a panic since this should never happen, but I can see arguments for both.

function/transform/transformation.go
```diff
@@ -58,27 +58,17 @@ func ApplyTransform(list api.SeriesList, transform transform, parameters []funct
...
-	return result, nil
-}
-
 // Integral integrates a series whose values are "X per millisecond" to estimate "total X so far"
 // if the series represents "X in this sampling interval" instead, then you should use transformCumulative.
 func Integral(values []float64, parameters []function.Value, scale float64) ([]float64, error) {
 	result := make([]float64, len(values))
 	integral := 0.0
 	for i := range values {
+		// Skip the 0th element since thats not technically part of our timerange
```
achow: Why is element 0 not part of our timerange?

function/transform/special.go
drcapulet: Since the condition below is only when the delta is zero, not sure how'd you write it to make sure that holds?

function/transform/special.go
drcapulet: I agree - but I say we decide on one type of error handling and I prefer passing around errors instead of causing a `panic()`

function/transform/transformation.go
drcapulet: The left most point has no "width" - probably easiest to graph in person. Essentially this makes sure that you can do `transform.derivative | transform.integral` and it'll work properly.

function/transform/special.go
drcapulet: True, but since these functions are so simple I left them independent. Lines of code would roughly be the same since you still need the range.

function/transform/special.go
achow: Oh.. you're actually changing the behavior of `rate`. Why? `rate` should just be `nonNegativeDerivative`.

I'm guessing this is to smooth out spikes from missing data? We should accomplish this using `nanKeepLast` or whatever we called it. This approximation totally breaks for counters anyways: `[1000000, nil, 1000030]` with 30s resolution becomes `[0, nil, 33334]`.

As a side note, if any component values are nil, the result should be 0. For derivative, we could make the result nil.

drcapulet: Talked in person, going to update the comment to make it clear that this is specifically for counters. Also need to look into our handling on NaNs

function/transform/special.go
achow: Can you add a comment here explaining that we make this approximation to handle when counters restart?

achow:
It might be better make grabbing extra data something like an `extraDataPoints` field in the execution context (or otherwise a higher concept) so it's easier to follow along where these changes occur. I can see things like moving average would want to leverage this (as well as that other PR you referenced).

Other than the comment on comments, LGTM.

function/transform/special.go
cchandler: Yes, one of the worst shortcomings of this thing's predecessors is inconsistent/unexposed handling of NaN/nil. I really like exposing that behavior. I don't like graphing systems that lie behind the scenes.

function/transform/special.go
cchandler: +1 I would also love a context object on these functions so you could append notes a la EQ's processing logs. It would be nice to expose how many times the approximation was used so you could dig into lossy data.

function/transform/special.go
cchandler: I like to think of panic() as go's version of runtime assert(). I think it's reasonable for cases where a system-generated value is clearly nonsense.

cchandler:
LGTM pending feedback.

PR #205, opened 2015-09-11
Adjust timerange with backend information

Nathan-Fenner:
This PR functions as a replacement for the functionally of https://github.com/square/metrics/pull/185.

This PR adds the `ChooseResolution()` method to the `TimeseriesStorageAPI` interface.

The method is provided with the user-entered timerange and the slot limit, and chooses a resolution for the query.

By giving the API control over what resolutions to choose, we can ensure that we use a resolution that is natively understood (e.g., don't use 7m resolutions if we only actually support 5m and 1hr).

Other changes:
* `Timerange`s now have `.StartTime()` and `.EndTime()` which return `time.Time` types
* `Timerange`s now have a `.Duration()` which returns a `time.Duration` value

### DIFF
--- a/api/timeseries_storage_api.go
+++ b/api/timeseries_storage_api.go
@@ -27,28 +27,6 @@ type TimeseriesStorageAPI interface {
 	FetchMultipleTimeseries(request FetchMultipleTimeseriesRequest) (SeriesList, error)
 }
 
-type ProfilingTimeseriesStorageAPI struct {
-	Profiler             *inspect.Profiler
-	TimeseriesStorageAPI TimeseriesStorageAPI
-}
-
-var _ TimeseriesStorageAPI = (*ProfilingTimeseriesStorageAPI)(nil)
-
-func (a ProfilingTimeseriesStorageAPI) ChooseResolution(requested Timerange, smallestResolution time.Duration) time.Duration {
-	defer a.Profiler.Record("timeseriesStorage.ChooseResolution")()
-	return a.TimeseriesStorageAPI.ChooseResolution(requested, smallestResolution)
-}
-
-func (a ProfilingTimeseriesStorageAPI) FetchSingleTimeseries(request FetchTimeseriesRequest) (Timeseries, error) {
-	defer a.Profiler.Record("timeseriesStorage.FetchSingleTimeseries")()
-	return a.TimeseriesStorageAPI.FetchSingleTimeseries(request)
-}
-
-func (a ProfilingTimeseriesStorageAPI) FetchMultipleTimeseries(request FetchMultipleTimeseriesRequest) (SeriesList, error) {
-	defer a.Profiler.Record("timeseriesStorage.FetchMultipleTimeseries")()
-	return a.TimeseriesStorageAPI.FetchMultipleTimeseries(request)
-}
-
 type FetchTimeseriesRequest struct {
 	Metric         TaggedMetric // metric to fetch.
 	SampleMethod   SampleMethod // up/downsampling behavior.
--- a/api/timeseries_storage_api.go
+++ b/api/timeseries_storage_api.go
@@ -22,7 +22,7 @@ import (
 )
 
 type TimeseriesStorageAPI interface {
-	ChooseResolution(requested Timerange, slotLimit int) time.Duration
+	ChooseResolution(requested Timerange, smallestResolution time.Duration) time.Duration
 	FetchSingleTimeseries(request FetchTimeseriesRequest) (Timeseries, error)
 	FetchMultipleTimeseries(request FetchMultipleTimeseriesRequest) (SeriesList, error)
 }
@@ -34,9 +34,9 @@ type ProfilingTimeseriesStorageAPI struct {
 
 var _ TimeseriesStorageAPI = (*ProfilingTimeseriesStorageAPI)(nil)
 
-func (a ProfilingTimeseriesStorageAPI) ChooseResolution(requested Timerange, slotLimit int) time.Duration {
+func (a ProfilingTimeseriesStorageAPI) ChooseResolution(requested Timerange, smallestResolution time.Duration) time.Duration {
 	defer a.Profiler.Record("timeseriesStorage.ChooseResolution")()
-	return a.TimeseriesStorageAPI.ChooseResolution(requested, slotLimit)
+	return a.TimeseriesStorageAPI.ChooseResolution(requested, smallestResolution)
 }
 
 func (a ProfilingTimeseriesStorageAPI) FetchSingleTimeseries(request FetchTimeseriesRequest) (Timeseries, error) {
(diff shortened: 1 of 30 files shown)


timeseries_storage/blueflood/blueflood.go
```diff
@@ -450,6 +450,44 @@ var samplerMap map[api.SampleMethod]sampler = map[api.SampleMethod]sampler{
 	},
 }
 
+// Blueflood will use the finest-grained resolution which doesn't exceed the slot limit.
+// Thus, if you request too many points, it will automatically downsample the data.
+func (b *Blueflood) AdjustTimerange(requested api.Timerange, slotLimit int) api.Timerange {
+	// In some cases, coarser-resolution data may have a shorter TTL.
+	// To accomodate these cases, it must be verified that the requested timerange will
+	// actually be present for the chosen resolution.
+	// TODO: figure out how to make this work with moving averages and timeshifts
```
drcapulet: This will also interact with my PR (#204) since we're doing time range altering there as well. Maybe we need to "adjust" the timerange before any functions are applied?

Nathan-Fenner: It's possibly a bit confusing, but `AdjustTimerange` is performed at the *beginning* of each `select` before evaluating the body, *not* before each fetch.

The reason for this is (1) none of the functions can tolerate working with data that has differing resolutions (aggregation and joins are much more complicated if the resolution is different in different parts of the data) and (2) no part of the code but the timeseries storage knows what resolutions are valid for which time periods.

Basically, when the user performs a query, the first thing that happens is that the resolution will be changed if the `TimeseriesStorageAPI` knows that the user-selected resolution won't work.

drcapulet: Gotcha - that makes sense. However it seems like the transforms altering the timerange would need to find a way to signal that to everything else so all the series are fetched at the correct resolution? (aka your window may not hit a boundary, but a moving average uses a larger window and may hit a limit).

Nathan-Fenner: Yes, that is a current limitation of this design. I think the goal was that eventually the TimeseriesStorageAPI would perform upsampling as needed (for example, if asked for 30s data that's 3 days old, it would instead query the 5m data, and interpolate to get faked 30s data). This upsampling would occur at each `.fetch()` 

Adjusting timerange mostly serves to prevent the `PointFetchLimit` from being exceeded. Otherwise, you'll get an annoying "too many points to fetch" which can be fixed really easily by just using a lower resolution. Right now, this limit ignores things that change the timerange (like moving average) since it only reads the timerange before it's put into the select expression.

api/timeseries_storage_api.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 type TimeseriesStorageAPI interface {
+	AdjustTimerange(requested Timerange, slotLimit int) Timerange
```
cchandler: Two things. 1) Structurally I don't think this belongs on the TimeseriesStorageAPI. 2) I want to modify the name so it makes more sense.

If this belongs on the top-level TimeseriesStorage than it needs to be at least universal enough that would apply to a second use-case. How would this work if the backing storage was a single RRD file? What would "adjustment" mean in that case? If it doesn't generalize then I would make it an exported method on the blueflood type, OR we create a slightly higher-abstraction that would generalize.

timeseries_storage/blueflood/blueflood.go
cchandler: Let's work through some examples together when we get some time. Apropos of my other remark, I don't like time range adjustment as a top-level concept in the API. At least at this point. But what it sounds like it's implying is that function evaluation context is what carries this time frame in addition to some kind of deeper response object so the underlying implementation can give hints back. If that were the case, it would effectively be a pre-validation step that just happens to either return an error, or support a flag to auto-adjust ranges.

Secondly, and semi-disconnected, I'm fairly strongly opposed to data interpolation at this point unless we have a very reference implementation and a good way to signal to the user that we've synthesized data. It's possible to work with variable resolution time ranges and I'd be more apt to start thinking about how we would integrate something like that now, before we go a lot farther with this design.

api/timeseries_storage_api.go
Nathan-Fenner: Right now I think the method name is certainly misleading. Having said that, it's not clear where else it should go- because the only place in the code that can know the internals of the API is the API implementation itself.

The specification for `AdjustTimerange` has the following requirements:

The resulting `Timerange` should:

1. Have (approximately) the same `Start` and `End` as the input (subject to changes due to resolution snapping)
2. Have a resolution *which the backend supports*
3. Have fewer slots than the specified `slotLimit`

The purpose of (1) is so that the user isn't shown a timerange that's completely different from what they asked for.

The purpose of (2) is so that the data can be queried without requiring up- or down-sampling. Otherwise, when the time to fetch the data occurs, the backend will find that it *needs* to sample the data at a new resolution.

The purpose of (3) is because if the result has too many slots, the query is rejected immediately, and nothing is evaluated (the user just gets back an error message).

The problem is essentially that (2) requires knowledge of the supported resolutions of the API, which only the API itself knows. Moreover, a list of resolutions isn't enough, because based on their TTLs or other characteristics, a given resolution might not be available for the specified `Timerange`.

One possibility would be instead of asking for a `Timerange`, just ask for a `resolution` (either in milliseconds or a `time.Duration`). Then the generic re-snapping can be performed at the call site, instead of inside the implementation.

Nathan-Fenner:
A general implementation should just choose the coarsest-grained resolution which is no finer than the user-specified resolution, while also being valid for that API, and being within the slot limit.

Nathan-Fenner:
As an example, if I'm implementing an API and my backend supports the resolutions 10m, 20m, 1h, 10hr, and their TTLs are (5d, 3d, 10d, forever) and I'm given a timerange (-4d, now, resolution: 15m) then I should return (-4d, now, resolution: 1h) since the 20m data won't be available at -4d.

Nathan-Fenner:
To make the purpose more clear and connected to what `TimeseriesStorageAPI`s do, the method has been renamed to `ChooseResolution`, which returns a `time.Duration` value.

The API should produce a resolution which it can support, and which will prevent the slot limit from being exceeded.

cchandler:
LGTM

Nathan-Fenner:
To make it easier for implementers of the `TimeseriesStorageAPI`, instead of being given a `slotLimit` that they aren't allowed to go over, they're given a `smallestResolution time.Duration` which is the smallest allowable duration that they could choose.

There's likely a small cost to precision. If the query is very close to the slot limit for a given resolution, the estimator is somewhat conservative and therefore may lead to necessary reductions in resolution. However, there should be a strong guarantee that the slot limit will never be exceeded, so long as the minimum resolution constraint is obeyed.

PR #206, opened 2015-09-11
Fix Profilers - Remove *Profiler from Contexts

Nathan-Fenner:
This PR fixes profiling in MQE. See #207 as well, which offers a different solution. 

This change moves the `*Profiler` pointers out of the `ExecutionContext`s, `EvaluationContext`s, and the `FetchSingle/MultipleRequest`s. 

Instead, the `*inspect.Profiler` pointer is placed exclusively in the `ProfilingMetadataAPI`, the `ProfilingTimeseriesStorageAPI`, and the `ProfilingCommand`.

However, because of this, it is not possible to profile individual `FetchSingleSeries` calls, since the `ProfilingTimeseriesStorageAPI` isn't able to capture these calls (single the `blueflood` object explicitly invokes the methods on itself).

In general, this might not be a problem (for example, a backend which uses batched requests wouldn't have any meaningful single-series requests to profile).

PR #207 goes in a different direction, instead getting rid of the `ProfilingAPI` types and exclusively passing the `*inspect.Profiler` pointer around in the contexts.

### DIFF
--- a/api/timeseries_storage_api.go
+++ b/api/timeseries_storage_api.go
@@ -48,7 +48,6 @@ type FetchTimeseriesRequest struct {
 	Timerange      Timerange    // time range to fetch data from.
 	MetricMetadata MetricMetadataAPI
 	Cancellable    Cancellable
-	Profiler       *inspect.Profiler
 }
 
 type FetchMultipleTimeseriesRequest struct {
@@ -57,7 +56,6 @@ type FetchMultipleTimeseriesRequest struct {
 	Timerange      Timerange
 	MetricMetadata MetricMetadataAPI
 	Cancellable    Cancellable
-	Profiler       *inspect.Profiler
 }
 
 type TimeseriesStorageErrorCode int
@@ -110,7 +108,6 @@ func (r FetchMultipleTimeseriesRequest) ToSingle() []FetchTimeseriesRequest {
 			Cancellable:    r.Cancellable,
 			SampleMethod:   r.SampleMethod,
 			Timerange:      r.Timerange,
-			Profiler:       r.Profiler,
 		}
 		fetchSingleRequests = append(fetchSingleRequests, request)
 	}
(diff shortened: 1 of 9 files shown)


cchandler:
I think between this and #207 I'm leaning toward this one. My one concern would be that we lose profiling information from within the query parser directly. Have we ever had performance issues in there? My guess is no, but I want to make sure we've considered it.

cchandler:
Actually, maybe the right answer is a hybrid between #206 and #207. In an ideal world a request would come in which would imply a 'context' that could accumulate profiling data. I don't necessarily think there was a problem with the *context objects carrying a profiler since it made it easy to propagate. Then, if the query builder creates FetchSingle/Multiple requests it could propagate the same profiler to those and it would be up to the underlying implementation to signal profiling events. Then by the time everything is done the final request would have an accurate + complete profiling tree. Does that make sense?

Nathan-Fenner:
We've never actually had profiling information on the parser- so I can't say for sure that it was never slow, but it has never seemed to be a problem either. (PEG allows linear time parsing, so it's unlikely that there are any pathological inputs).

Nathan-Fenner:
@cchandler what you described is essentially #207, where the `*Profiler` is carried first in the `ExecutionContext`, then in the `EvaluationContext`, and then put into the `FetchMultipleSeriesRequest` and then (if needed) into the `FetchSingleSeriesRequest`.

The only real disadvantage towards #207 is that every API has to do its own profiling to be complete and consistent, because something that wraps up the Blueflood API can't record the calls it makes to its own methods.

cchandler:
@Nathan-Fenner Yeah, you're right. I thought about it more at the gym last night :-). I'm more strongly in favor of #207 but instead of making that second parameter on the MetricMetadataAPI a profiler, something like a MetricMetadataAPIContext that happens to have a profiler on it. That way we have an auxiliary struct we can pass around with ancillary capability on it without polluting the API excessively. That will also let an underlying implementation access a profiler if it needed it.

Nathan-Fenner:
#206 is closing in favor of #207

PR #207, opened 2015-09-12
Fix Profiler - Delete ProfilingAPI types

Nathan-Fenner:
This PR fixes profiling in MQE. See #206 as well, which offers a different solution.

In this version, the `*inspect.Profiler` pointers are carried by the `ExecutionContext` and the `EvaluationContext`, and added to the `FetchTimeseriesRequest` and `FetchMultipleTimeseriesRequest` when they are created. The `*Profiler` for the `MetricMetadataAPI` is placed into a `MetricMetadataAPIContext` which is added as an argument to every call. Currently, it's simplest created by putting the `*Profiler` from the execution context or the evaluation context into it.

Each MetadataAPI and TimeseriesStorageAPI implementation is responsible for performing its own profiling- there's no longer any kind of `ProfilingFooAPI`. 

~~The main advantage of #207 over #206 is that you're able to get profiling data from `.FetchSingleTimeseries` in addition to the aggregate time total from `.FetchMultipleTimeseries` This disadvantage of this design is that each new API (including the mocks, etc.) are required to implement their own profiling.~~

#206 has been closed in favor of #207.

### DIFF
--- a/testing_support/mocks/api.go
+++ b/testing_support/mocks/api.go
@@ -34,8 +34,8 @@ type FakeMetricMetadataAPI struct {
 
 var _ api.MetricMetadataAPI = (*FakeMetricMetadataAPI)(nil)
 
-func NewFakeMetricMetadataAPI() FakeMetricMetadataAPI {
-	return FakeMetricMetadataAPI{
+func NewFakeMetricMetadataAPI() *FakeMetricMetadataAPI {
+	return &FakeMetricMetadataAPI{
 		metricTagSets: make(map[api.MetricKey][]api.TagSet),
 		metricsForTags: make(map[struct {
 			key   string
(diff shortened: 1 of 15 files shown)


testing_support/mocks/api.go
```diff
@@ -33,8 +34,8 @@ type FakeMetricMetadataAPI struct {
 
 var _ api.MetricMetadataAPI = (*FakeMetricMetadataAPI)(nil)
 
-func NewFakeMetricMetadataAPI() FakeMetricMetadataAPI {
-	return FakeMetricMetadataAPI{
+func NewFakeMetricMetadataAPI() *FakeMetricMetadataAPI {
```
Nathan-Fenner: Returning a pointer is more useful than the value, since the pointer implements the `MetadataAPI` interface. Everywhere that called this method had to use `&myAPI` everywhere, which was unnecessary.

cchandler:
Good direction. LGTM.

PR #208, opened 2015-09-16
Kuzmik testing ci

kuzmik:
This addresses two separate timeouts that I saw happen in CI builds yesterday. I ran over 80 builds and only had 2 show up.

I also ran the tests in a loop on my laptop, and in 542 runs there were 0 timeouts; so it is clearly something a bit weird with travis. Hopefully this does something to alleviate SOME of the pain.


achow:
LGTM

PR #209, opened 2015-09-18
Fixes from yesterdays outage

kuzmik:
* schema now uses replication factor of 3
* using quorum on reads is unnecessary

### DIFF
--- a/metric_metadata/cassandra/cassandra.go
+++ b/metric_metadata/cassandra/cassandra.go
@@ -35,6 +35,7 @@ type CassandraMetricMetadataConfig struct {
 // NewCassandraMetricMetadataAPI creates a new instance of API from the given configuration.
 func NewCassandraMetricMetadataAPI(config CassandraMetricMetadataConfig) (api.MetricMetadataAPI, error) {
 	clusterConfig := gocql.NewCluster()
+	clusterConfig.Consistency = gocql.One
 	clusterConfig.Hosts = config.Hosts
 	clusterConfig.Keyspace = config.Keyspace
 	clusterConfig.Timeout = time.Second * 30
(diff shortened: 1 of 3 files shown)


metric_metadata/cassandra/cassandra.go
```diff
@@ -35,6 +35,7 @@ type CassandraMetricMetadataConfig struct {
 // NewCassandraMetricMetadataAPI creates a new instance of API from the given configuration.
 func NewCassandraMetricMetadataAPI(config CassandraMetricMetadataConfig) (api.MetricMetadataAPI, error) {
 	clusterConfig := gocql.NewCluster()
+	clusterConfig.Consistency = gocql.One
```
jeeyoungk: i think we should use quorum, with replication_factor set to 3 (or even 5!) we should be safe.

kuzmik: Why do we need quorum for reads?

On Friday, September 18, 2015, Jeeyoung Kim <notifications@github.com>
wrote:

> In metric_metadata/cassandra/cassandra.go
> <https://github.com/square/metrics/pull/209#discussion_r39916418>:
>
> > @@ -35,6 +35,7 @@ type CassandraMetricMetadataConfig struct {
> >  // NewCassandraMetricMetadataAPI creates a new instance of API from the given configuration.
> >  func NewCassandraMetricMetadataAPI(config CassandraMetricMetadataConfig) (api.MetricMetadataAPI, error) {
> >  	clusterConfig := gocql.NewCluster()
> > +	clusterConfig.Consistency = gocql.One
>
> i think we should use quorum, with replication_factor set to 3 (or even
> 5!) we should be safe.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/square/metrics/pull/209/files#r39916418>.
>


-- 
Typos and succinctness are courtesy of my iPhone.

cchandler:
LGTM

PR #210, opened 2015-09-20
Ruletester Cleanup / Usability Improvements

Nathan-Fenner:
While trying out the new ruletester, I ran into a few problems that were pretty straightforward to solve:

* When a flag is missing, say what flag it is: namely `-metrics-file` and `-rule-path`
* Don't ask for a complete configuration file (doesn't need `-config-file`)- only the conversion rule path was being used, so the rest of the config was extraneous
* Increase `runtime.GOMAXPROCS` to `runtime.NumCPU()` so that it runs fully in parallel - this increased speed slightly on my machine

I also just cleaned up the commented code and some unused functions, and added error checking in a handful of places.

The batching of metrics obscured what was going on in the code a little bit- so I got rid of it. Running the program a few times on my machine took ~65s (real/clock time) either way.

To help find buggy conversion rules, metrics which fail to be reversed are printed after the ruletester runs. Additionally, rules are now checked for successful conversion back into the original graphite, and these are also printed if they differ from the expected original.

### DIFF
--- a/main/ruletester/ruletester.go
+++ b/main/ruletester/ruletester.go
@@ -24,7 +24,6 @@ import (
 	"encoding/gob"
 	"flag"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


main/ruletester/ruletester.go
```diff
@@ -66,22 +62,21 @@ func ReadMetricsFile(file string) ([]string, error) {
...
-	err = d.Decode(&result)
+	// Store the result of the decode in this map:
+	result := map[string]int{}
+	err = gob.NewDecoder(r).Decode(&result)
 	if err != nil {
 		return nil, err
 	}
 
-	strings := []string{}
+	strings := make([]string, 0, len(result))
```
cchandler: Let's come up with consistent style. This one keeps coming up. For an empty slice I'm leaning toward `[]string{}`. @achow ?

main/ruletester/ruletester.go
```diff
@@ -89,171 +84,146 @@ func ReadMetricsFile(file string) ([]string, error) {
 }
 
 func main() {
+	runtime.GOMAXPROCS(runtime.NumCPU())
```
cchandler: This has always been configurable via the environment. Let's only override it if the environment variable isn't set.

main/ruletester/ruletester.go
achow: I like `[]string{}` since it's pretty concise. `make()` with capacity is a bit of a micro-optimization, but I'm ok with it if we already know how big the slice is going to be.

cchandler:
LGTM pending feedback.

PR #211, opened 2015-09-22
Add logging for possible failure cases

Nathan-Fenner:
Investigating a bug in MQE suggests that there may be a problem with some part of the graphite => tagged-metric conversion. There is a buggy tendency to introduce NULL and other illegal characters into the strings. This logging should make it straightforward to tell if there is a problem.


syamp:
lgtm

PR #212, opened 2015-09-23
Add an externalized optimizer.

cchandler:
The optimize functions are external to both query/function because
go forbids import loops and the capabilities have to be available
in both. Further, I'd like this to be as implementation agnostic
as possible since it's technically for query optimization, so I
opted to put it in its own package for now rather than push it into
the underlying metricmetadata API

### DIFF
--- a/query/command.go
+++ b/query/command.go
@@ -23,18 +23,20 @@ import (
 	"github.com/square/metrics/function"
 	"github.com/square/metrics/function/registry"
 	"github.com/square/metrics/inspect"
+	"github.com/square/metrics/optimize"
 	"github.com/square/metrics/query/natural_sort"
 )
 
 // ExecutionContext is the context supplied when invoking a command.
 type ExecutionContext struct {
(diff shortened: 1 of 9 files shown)


query/command.go
```diff
@@ -73,9 +75,17 @@ type SelectCommand struct {
...
 // Execute returns the list of tags satisfying the provided predicate.
 func (cmd *DescribeCommand) Execute(context ExecutionContext) (interface{}, error) {
-	tagsets, _ := context.MetricMetadataAPI.GetAllTags(cmd.metricName, api.MetricMetadataAPIContext{
-		Profiler: context.Profiler,
-	})
+
+	// We generate a simple update function that closes around the profiler
+	// so if we do have a cache miss it's correctly reported on this request.
+	updateFunction := func() ([]api.TagSet, error) {
+		tagsets, err := context.MetricMetadataAPI.GetAllTags(cmd.metricName, api.MetricMetadataAPIContext{
```
drcapulet: Why not just inline return?

drcapulet:
:+1:

PR #213, opened 2015-09-24
Fix copy

cchandler:
I screwed up copy semantics in the move. The code was closing around the wrong variable and was thus always the original.


drcapulet:
:+1:

PR #215, opened 2015-09-25
Cleanup for select queries and fixes to tests and mocks

Nathan-Fenner:
Previously, when a timeout was specified, `select` actually behaved slightly differently than when it was not. In particular, it apparently bypassed checks that the resulting expressions were serieslists.

This PR unifies the approach taken when a timeout is present and when it is absent. Because of this, when the timeout is not specified, one additional goroutine is spawned for the duration of the request than previously existed.

Also, there was a small potential for a very gradual leak. Go does *not* garbage collect goroutines, and sending on a channel with default capacity (0) will block until a receiver comes along. In the event of a timeout, the goroutine that was launched to evaluate the expression was never able to finish, so it would hang onto the memory for its channels and parameters forever.

The mock API didn't actually behave the same as the Cassandra API for non-existent tags; some tests were sorta expecting this, but for whatever reason the other mismatches obscured this. Now, the mock metadata API will error when asked for the tags of a metric which does not exist.

Relating to all of these challenges- the tests relating to `select` queries were broken in a few ways, which has been corrected.

### DIFF
--- a/query/command.go
+++ b/query/command.go
@@ -208,7 +208,7 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (interface{}, error)
 		OptimizationConfiguration: context.OptimizationConfiguration,
 	}
 
-	timeout := chan time.Time(nil)
+	timeout := <-chan time.Time(nil)
 	if hasTimeout {
 		// A nil channel will just block forever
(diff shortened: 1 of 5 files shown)


cchandler:
LGTM

PR #216, opened 2015-09-27
Fix typo in timerange selection

Nathan-Fenner:
A typo was introduced in the timerange selection code for executing `select` commands.

This caused MQE to ignore the slot limit and the API-specified best resolution.

This PR fixes the typo and renames the involved variables to improve clarity.


cchandler:
LGTM

PR #217, opened 2015-09-27
Fix race condition in metric tag caching

Nathan-Fenner:
This PR locks the cache mutex before reading from the `metricsKeyToTagCache` map.

Maps in Go are not threadsafe. It is only safe for there to be *one writer* **or** *many readers* **but not both**.

If a goroutine writes to the map while another goroutine is reading it, garbage may be read from it.


Nathan-Fenner:
This is likely the source of the invalid metrics found in production.

cchandler:
LGTM, but I'm skeptical it's connected to the invalid metrics. Those metrics, included the ones I see in my test lab predate this change.

Nathan-Fenner:
Yeah- with some reflection I realized that the race condition actually happening (often) seems very unlikely.

Issue #214, opened 2015-09-25
Describe Should Work With Queries


drcapulet:
It'd be awesome to be able to do this:

```
describe `a.b.c` | transform.rate
```

Nathan-Fenner:
It would probably be more straightforward to instead add descriptions to all `select` queries. As long as you're only interested in the tags from the resulting series rather than the queried series (so, for example, `metric | aggregate.sum(group by datacenter) from -2hr to now` would only have datacenter tags) then this would be very simple the implement.

The main challenge in adding this information to `select` is that in the JSON representation, `select` is just a list- so there's no place for us to put the extra data without changing the API.

Also, would it be better to use the existing `describe` behavior, which produces a `map[Key][]Value`, or the more direct behavior, which produces a `[]map[Key]Value`?

Nathan-Fenner:
It's now possible to get description information from a `select` statement, through the `metadata` field in the JSON response: `response.metadata.description` from a select statement has the same form as the body of a describe statement.

PR #218, opened 2015-09-28
Add metadata to command execution results

Nathan-Fenner:
As a way to address issue https://github.com/square/metrics/issues/214 adds a `metadata` field to the JSON response from MQE.

The metadata map can contain information that is secondary to the result itself. Implemented here are descriptions for `select` queries (the tagsets of the resulting series are combined) and profiling information. (To maintain reverse compatibility, the original `.profile` field is still available in the JSON response, but now it is *also* available through `.metadata.profile` which will allow us to eventually move to a state where profiling information is not quite as privileged in that sense).

### DIFF
--- a/main/query/query.go
+++ b/main/query/query.go
@@ -76,7 +76,7 @@ func main() {
 			fmt.Println("execution error:", err.Error())
 			continue
 		}
-		encoded, err := json.MarshalIndent(result, "", "  ")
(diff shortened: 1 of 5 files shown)


cchandler:
Seems reasonable to me. LGTM. This will help me with exposing the EvaluationNotes in my PR :-)

cchandler:
I'm going to go ahead and merge this because I want to add some metadata :-)

PR #219, opened 2015-09-30
Context notes on function evaluation

cchandler:
This modifies the usage of EvaluationContext to include EvaluationNotes during function evaluation. Now that it passes a pointer around please be careful with copy semantics :-).

Example:
The non-negative derivative functionality now creates a context note like this:
  ctx.AddNote(fmt.Sprintf("Rate(%v): The underlying counter reset between %f, %f\n", series.TagSet, values[i-1], values[i]))
This passes information back to the UI that wouldn't be obvious from the processed numbers alone. It will include things like strange rounding, empty data buckets, unexpected nils, interpolation, whatever.

It also changes the signature of the transform function type to work directly on api.Timeseries instead of a simplified, contextless []float64.

A few TODOs were added that I'll address later this week. Tests were only modified to work with the new code, no numbers were modified.

### DIFF
--- a/function/expression.go
+++ b/function/expression.go
@@ -28,6 +28,7 @@ type EvaluationContext struct {
 	Profiler                  *inspect.Profiler // A profiler pointer
 	OptimizationConfiguration *optimize.OptimizationConfiguration
 	EvaluationNotes           []string //Debug + numerical notes that can be added during evaluation
+	invalid                   bool     // Because these can be copied, it's best to mark a no-longer used context as dead
 }
 
 type Registry interface {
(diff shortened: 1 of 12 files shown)


drcapulet:
I'd check on `newDerivativeBasedTransform` since we edit the context there, also may want to check the evaluation context coming back for notes and copy them.

cchandler:
I've updated the `newDerivativeBasedTransform` to use the new Copy + Invalidate operations. Same goes for the moving average.

drcapulet:
:+1:

PR #220, opened 2015-09-30
Add notes to metadata return and improve profiling result for blueflood

cchandler:
This will add evaluation notes to the metadata response. It also now uniquely identifies FetchSingle requests in the response.
metrics_query_engine

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -19,6 +19,10 @@
 
 package api
 
+import (
+	"fmt"
+)
+
 // list of data types throughout the code.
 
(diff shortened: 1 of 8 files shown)


drcapulet:
:+1:

cchandler:
I opted to go with @Nathan-Fenner 's suggestion and just move the new info there to prevent swamping the metrics infra upstream. Just in case. It includes the info in the UI for hover. I also went ahead and added a USR2 signal handler on the ui tool so we can get a goroutine dump on demand (turned out to be very useful tracking some stuff down today).

PR #221, opened 2015-09-30
Add some new features to the ruletester

Nathan-Fenner:
Adding a bit more information to the ruletester makes it easier to diagnose some problems.

The report files contain the full metric name from source, instead of the regex, and also list the file that the conversion rule comes from.


cchandler:
:+1:

PR #222, opened 2015-10-01
Include raw payload

cchandler:
I've added a flag on the query string called 'raw'. If set to true it will base64 all the original payload results from the underlying TimeseriesStorageAPI onto the response body. Testing this week has revealed several cases in which I wished I had direct access to the underlying data.

I'll add some tests in the morning.

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -95,8 +95,9 @@ func bodyResponse(writer http.ResponseWriter, request *http.Request, response re
 // -----------------
 
 type queryForm struct {
-	input   string // query to execute.
-	profile bool   // if true, then profile information will be exposed to the user.
+	input      string // query to execute.
+	profile    bool   // if true, then profile information will be exposed to the user.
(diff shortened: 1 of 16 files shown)


cchandler:
PS The MarshalJSON() on the Timeseries object is gross and needs to go.

drcapulet:
:+1:, though it'd be awesome if we could trigger it via query params.

ui/ui.go
```diff
@@ -163,7 +165,11 @@ func (q queryHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reques
 	}
 
 	cmd, profiler := query.NewProfilingCommand(cmd)
-	result, err := cmd.Execute(q.context)
+	ctx := q.context
+	ctx.UserSpecifiableConfig = api.UserSpecifiableConfig{
```
drcapulet: Derp, missed this.

PR #223, opened 2015-10-02
Add `does_not_match` field to conversion rules

Nathan-Fenner:
This allows you to exclude certain metrics from a conversion rule by rejecting the match if the result's tag values match their `do_not_match` field.

Golang does not support negative lookahead in its regular expression (in order to keep them more regular, as well as to ensure O(n) matching) and therefore it can be very difficult to intentionally exclude certain metrics from being matched.

It used to be that metrics were just matched in order, so that you could exclude from one rule by matching with another. However, because of danger of this introducing subtle bugs when one rule was changed, the behavior was changed so that order must not matter, and all metrics should match at most one rule.

This change makes it possible to easily exclude metrics from a particular rule, while keeping it orthogonal to all others- so that if another rule is removed or changed, it won't suddenly start capturing everything that the old one had.

TODO: check this when reversing metrics too (this is less likely to be useful, but for completeness it's not a bad idea)

### DIFF
--- a/util/rules.go
+++ b/util/rules.go
@@ -32,7 +32,7 @@ type RawRule struct {
 	Pattern          string            `yaml:"pattern"`
 	MetricKeyPattern string            `yaml:"metric_key"`
 	Regex            map[string]string `yaml:"regex,omitempty"`
-	DoesNotMatch     map[string]string `yaml:"does_not_match,omitempty"`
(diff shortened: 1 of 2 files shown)


syamp:
I like this feature!

syamp:
@cchandler

drcapulet:
:+1:

PR #224, opened 2015-10-03
Implements XOR-based float64 slice compression

cchandler:
I took a few hours and wrote a sample implementation of the streaming XOR-based compression algorithm that Facebook vaguely discusses in their Gorilla paper. Initial experimentation reveals that it has pretty good performance characteristics. The code is a little rough in parts, but it has several test vectors.

A possible future direction (especially for larger slabs) would be to implement a different version that uses the differential-finite-context-method predictor enhancements discussed in "Fast Lossless Compression of Scientific Floating-Point Data" by Paruj Ratanaworabhan, Jian Ke, and Martin Burtscher. If I'm reading it right it sounds like they save the prediction table (at least in part) but it might lead to better space saving overall.

### DIFF
--- a/compress/compress.go
+++ b/compress/compress.go
@@ -16,181 +16,19 @@ package compress
 import (
 	"bytes"
 	"encoding/binary"
-	"fmt"
 	"math"
 )
-
-const MaxUint32 = ^uint32(0)
 
 type CompressionBuffer struct {
 	buffer                  *bytes.Buffer
 	current                 uint64
 	position                uint32
-	leadingZeroWindowLength uint64
-	lengthOfWindow          uint64
+	leadingZeroWindowLength uint32
+	lengthOfWindow          uint32
 
 	finalized   bool
 	firstPass   bool
 	previousXOR uint64 //The last float we compressed
-}
-
-type DecompressionBuffer struct {
-	data             []byte
-	current          byte
-	currentByteIndex uint32
-
-	leadingZeroWindowLength uint64
-	lengthOfWindow          uint64
-	position                uint32
-	eof                     bool
-	expectedSize            int
-}
-
-func NewDecompressionBuffer(data []byte, expectedSize int) DecompressionBuffer {
-	dbuf := DecompressionBuffer{
-		data:         data,
-		position:     uint32(7), // Start reading from the "left"
-		eof:          false,
-		expectedSize: expectedSize,
-	}
-
-	if len(data) <= 8 {
-		//Tiny input.
-		dbuf.eof = true
-	} else {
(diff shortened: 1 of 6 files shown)


compress/compress.go
```diff
@@ -0,0 +1,477 @@
...
+
+func (c *CompressionBuffer) writeMeaningfulPartialXOR(length uint64, meaningful uint64) {
+	i := uint32(length)
+	for i != MaxUint32 {
+		bit := (meaningful >> i) & 1
+		if bit == 1 {
+			c.writeOne()
+		} else {
+			c.writeZero()
+		}
```
Nathan-Fenner: This same sort of code appears several times. I think it would make sense to make a function:

```
func (c *CompressionBuffer) writeBits(value uint64, n uint32) {
    i := n
    for i != MaxUint32 {
        bit := (value >> i) & 1
        if bit == 1 {
            c.writeOne()
        } else {
            c.writeZero()
        }
        i--
    }
}
```

which will hopefully make it easier to understand (by abstracting away some of these lower-level details).

Actually, it might also be helpful to have

```
func (c *CompressionBuffer) writeBit(bit bool) {
    if bit {
        c.writeOne()
    } else {
        c.writeZero()
    }
}
```

jeeyoungk:
this is a pretty nifty PR.

compress/compress.go
cchandler: That highlights an interesting choice of style. I find in implementing scientific/numerical algorithms that I favor more overt duplication of code because it lets me read the paper and implement it with a common language for larger-block pieces. For instance, in this case the writePartialXOR I can effectively read it in one go.

As an exercise, since you said "hopefully make it easier to understand", try checking out the branch and making the modifications and see if it helps with understanding. It's a fairly simple algorithm, but understanding it code-first is going to be tricky.

Nathan-Fenner: I went ahead and did try changing the style to see if it makes it easier to understand. It definitely helped me figure out exactly how it worked- although that probably has more to do with rewriting bits of it than how it actually turned out.



Code: https://github.com/square/metrics/compare/master...nathanfenner-style-experiments?expand=1#files_bucket

Diff: https://github.com/square/metrics/compare/cchandler-xor...nathanfenner-style-experiments?expand=1#files_bucket

cchandler: It's a good breakout. Feel free to push the changes onto my branch and merge :-)

Nathan-Fenner: Alright. As a sanity check, I also threw in a test to make sure compressing actually makes the data smaller.

PR #226, opened 2015-10-09
Add gaps in cumulative transform results

Nathan-Fenner:
This change is tentative. It changes the behavior of `transform.integral`, `transform.cumulative`, and `transform.moving_average` in the following way:

Values that are missing in the input and missing in the output. The results are otherwise the same (missing values treated as 0 for integrals/sums, and treated as missing for transform.moving_average).

The result is that the `transform.integral` function will no longer hide missing data- the missing data will visible as gaps in the result.

To obtain the previous behavior, instead use something like `transform.nan_keep_last(transform.integral(series))` which will fill the empty segments with the flat values.

This doesn't affect aggregations or filters. However, series produced with `transform.(integral|cumulative|moving_average)` may be sorted in a different order by `filter.*` (since the input has changed, the sort order may change).

A concrete example:

Consider input of the form `transform.cumulative([3, NaN, 4, NaN, NaN, 2, NaN])`

Old behavior: `[3, 3, 7, 7, 7, 9, 9]`

New behavior: `[3, NaN, 7, NaN, NaN, 9, NaN]`

### DIFF
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -118,7 +118,7 @@ var MovingAverage = function.MetricFunction{
 				}
 				// Numerical error could (possibly) cause count == 0 but sum != 0.
 				if i-limit+1 >= 0 {
-					if count == 0 {
+					if count == 0 || math.IsNaN(series.Values[i]) {
 						results[i-limit+1] = math.NaN()
 					} else {
 						results[i-limit+1] = sum / float64(count)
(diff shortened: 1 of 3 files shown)


syamp:
I see the reasoning behind this behavior change but I am not sure if the new result is any more correct than previous.

Nathan-Fenner:
Yes, I don't know whether this behavior is actually preferable.

Until a specific use-case can be found where the change would be beneficial, it's probably better to close this.

PR #228, opened 2015-10-12
Add errors to AddMetrics calls for Cassandra API

Nathan-Fenner:
Previously, calls to `AddMetrics` in the Cassandra API instances always returned `nil` to indicate error, even if there was an error. This was because they didn't actually check errors returned by query execution, for example.

This pull request adds error checking. It also fixes a very small space leak which would cause some goroutines to hang forever in the event of a query timeout.

### DIFF
--- a/metric_metadata/cassandra/cassandra.go
+++ b/metric_metadata/cassandra/cassandra.go
@@ -63,8 +63,7 @@ func (a *CassandraMetricMetadataAPI) AddMetric(metric api.TaggedMetric, context 
 
 func (a *CassandraMetricMetadataAPI) AddMetrics(metrics []api.TaggedMetric, context api.MetricMetadataAPIContext) error {
 	defer context.Profiler.Record("Cassandra AddMetrics")()
-	a.db.AddMetricNames(metrics)
-	return nil
+	return a.db.AddMetricNames(metrics)
 }
 
 func (a *CassandraMetricMetadataAPI) GetAllTags(metricKey api.MetricKey, context api.MetricMetadataAPIContext) ([]api.TagSet, error) {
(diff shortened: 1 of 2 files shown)


metric_metadata/cassandra/cassandra.go
```diff
@@ -130,42 +129,33 @@ func (db *cassandraDatabase) AddMetricNames(metrics []api.TaggedMetric) error {
 	queryInsert := "INSERT INTO metric_names (metric_key, tag_set) VALUES (?, ?)"
 	queryUpdate := "UPDATE metric_name_set SET metric_names = metric_names + ? WHERE shard = ?"
 
-	c := make(chan *gocql.Query, 10)
-	done := make(chan bool)
-	go func() {
```
Nathan-Fenner: It's not clear to me why a separate goroutine was spawned here- it's simpler to execute the queries synchronously.

Unless I'm mistaken, preparing the queries (calling `Bind`, but not `Exec`) should take almost no time, so trying to do this concurrently with execution shouldn't have any improvement for performance.

timeseries_storage/blueflood/blueflood.go
```diff
@@ -333,8 +333,8 @@ func (b *Blueflood) constructURL(
 // fetches from the backend. on error, it returns an instance of api.TimeseriesStorageError
 func (b *Blueflood) fetch(request api.FetchTimeseriesRequest, queryUrl *url.URL) (queryResponse, []byte, error) {
 	log.Debugf("Blueflood fetch: %s", queryUrl.String())
-	success := make(chan queryResponse)
-	failure := make(chan error)
```
Nathan-Fenner: In the event of a timeout, the goroutine spawned to perform the fetch will wait for someone to receive from `success` or `failure`. By giving them capacity, it won't have to wait for someone on the other end.

Otherwise, this will create garbage that Go's garbage collector won't ever touch- the goroutine itself will become eternally frozen.

metric_metadata/cassandra/cassandra.go
cchandler: Entirely possible. This is mostly a failing of the go driver for cassandra. The Java/C# driver handles this correctly for bulk-inserts.

cchandler:
LGTM.

PR #230, opened 2015-10-20
Bugfix for transform.rate

Nathan-Fenner:
Fixes a bug when using `transform.rate`.

Downsampling can cause sudden drops to be smeared out over several slots in the timeseries. The result was mis-reporting many resets over the interval of time that the downsampling occurred.

Since the resets were only partial, the values looked really, really large for long-running counters, and, consequently, caused enormous spikes.

This change requires that the next point show an increase (in addition to decreasing from the previous) to be considered a reset.

### DIFF
--- a/function/transform/special.go
+++ b/function/transform/special.go
@@ -194,6 +194,11 @@ func rate(ctx *function.EvaluationContext, series api.Timeseries, parameters []f
 		// Scaled difference
 		result[i-1] = (values[i] - values[i-1]) / scale
 		if result[i-1] < 0 {
+			result[i-1] = 0
+		}
+		if i+1 < len(values) && values[i-1] > values[i] && values[i] <= values[i+1] {
+			// Downsampling may cause a drop from 1000 to 0 to look like [1000, 500, 0] instead of [1000, 1001, 0].
+			// So we check the next, in addition to the previous.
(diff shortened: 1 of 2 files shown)


drcapulet:
Are we not able to tell MQE how we want to downsample via the query? Or can't `transform.rate` tell the underlying fetcher how it wants to sample? In this case, we want the extremes to correctly calculate counter wraps.

Nathan-Fenner:
There are a couple sources of downsampling that expose the problem:
* Rollups (for mean-sampled data, the mean will be smeared out)
* Data within 1-day may have downsampling artifacts, even though it's not downsampled

(2) is the source of the actual issue we noticed first (but I'm surprised (1) didn't have more often). In order to fill in data that is missing for low-resolution data (since rollups haven't been performed yet), *all requests for data within the last 24 hours are resampled with full-resolution data.* In this case, even if the 5-minute has no issues, combining it with the full-resolution data may introduce some. The alternative is to have no data at all- so this is probably preferable.

Nathan-Fenner:
As to why `transform.rate` doesn't choose the sampling method- right now, none of the functions ever change the sampling method. If they did, this might be surprising for users, since they wouldn't be able to override it, and the change happens silently.

syamp:
lgtm

PR #232, opened 2015-10-21
Demo data and conversion rules

Nathan-Fenner:
This PR makes it easier to start up a local instance of MQE and start indexing metrics.

`/demo/emit` has a program which emits sample metric data over HTTP to a Blueflood instance and to a metric indexe.r

`/demo/ingest` has a program which collects metric metdata over an HTTP endpoint and stores the metsdata in a cassandra instance,

The `/demo/conversion_rules` has conversion rules for the demo metrics.

The `/demo/example-config.yaml` file has an example of a configuration file to start the MQE UI.

Detailed documentation for starting and using the server will be made available on the GitHub Wiki for this project.

@cchandler @drcapulet @syamp

### DIFF
--- a/demo/emit/mock.go
+++ b/demo/emit/mock.go
@@ -1,3 +1,17 @@
+// Copyright 2015 Square Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


demo/emit/mock.go
```diff
@@ -0,0 +1,209 @@
...
+			metric.Source.Advance(iteration)
+			err := reportMetric(client, metric.Source.Value(), metric.Metric)
+			if err != nil {
+				fmt.Printf("Error for metric %s:\n%s\n", metric.Metric, err.Error())
+			}
+		}
+	}
+
+	//
+	/*
```
cchandler: Did you want to leave the comment block?

Nathan-Fenner: This PR isn't really ready to be merged (I'll add a note) but I thought it was a good idea to share what I've done so far.

I will be getting rid of the comment block, once I add some more sample data.

Nathan-Fenner: It's been cleaned up and is actually ready for review now. I added configuration via flags and an example `config.yaml` file

PR #233, opened 2015-10-27
Holt-Winters Forecasting for Timeseries

Nathan-Fenner:
TODO: Test this. I haven't done any at all, besides compiling.

Adds the function `forecast.holt_winters(series, period, training_time, training_duration)` to MQE.

The function uses the multiplicative Holt-Winters model for time series, namely:

    f(t) = S(t) * (a + b*t)

Where S(t) is a periodic function of known period and mean 1.

There are other ways to implement forecasting. This way does the following:

1. You specify an interval of time (offset in time and duration) to train the model
1. The query argument is evaluated for that interval
1. The data is used to calculate a Holt-Winters model that best approximates that data
1. The model is used to fill in the current timerange with data

There is an adaptive method which uses exponential smoothing. The model is the same, only it uses an incremental prediction of the data rather than finding a best-fitting model to start. I will be implementing this version soon.

### DIFF
--- a/api/timerange.go
+++ b/api/timerange.go
@@ -143,6 +143,12 @@ func (tr Timerange) Shift(shift time.Duration) Timerange {
 	return tr.Snap()
 }
 
+// ExtendLength returns a timerange which whose length is set to the given amount
+func (tr Timerange) SelectLength(length time.Duration) Timerange {
+	tr.end = tr.start + int64(length/time.Millisecond)
+	return tr.Snap()
+}
+
 // Slots represent the total # of data points
 // Behavior is undefined when operating on an invalid Timerange. There's a
 // circular dependency here, but it all works out.
--- a/function/forecast/forecast.go
+++ b/function/forecast/forecast.go
@@ -14,9 +14,8 @@
 
 package forecast
 
-import (
-	"math"
-)
+import "fmt"
+import "math"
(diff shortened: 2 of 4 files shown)


cuzelac:
@cchandler @jeeyoungk - would you mind taking a look at this when you get a free moment?

function/forecast/utility.go
```diff
@@ -0,0 +1,97 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package forecast
+
+// Returns the unique integer r such that x == r (mod m) and 0 <= r < m
```
cchandler: I'm guessing you wanted this after the import.

cchandler:
At a first pass, it looks fine. Given the numerical nature of much of the coding, scrutinizing it via code review is mostly a useless endeavor. I'm fine merging it so we can engage in a more pragmatic exploration.

Nathan-Fenner:
Okay- I think the most dangerous effect is if there's a panic somewhere in the code which isn't caught.

PR #234, opened 2015-10-29
Add support for metric tag-key indexing

Nathan-Fenner:
In the transition from using `AddMetric` to `AddMetrics` in the Cassandra API, it was forgotten to include the tag-key indexing inserts.

This PR allows you to use the query `describe metrics where key = 'value'` when indexing with `.AddMetrics`.


drcapulet:
:+1:

PR #235, opened 2015-11-04
Disable rate approximations on counter reset

achow:
@drcapulet


drcapulet:
@achow this was already fixed in #230, we just need to deploy it.

drcapulet:
Closing out.

PR #238, opened 2015-11-11
Remove checks for future queries

Nathan-Fenner:
In anticipation of https://github.com/square/metrics/pull/233, removes the restriction that relative fetches (`from -1d to now`) must be in the past. So you can now write `from now to +3d`


syamp:
lgtm

PR #240, opened 2015-11-23
replace p-values with standard stddevs; fix bug in periodic anomaly function

Nathan-Fenner:
The forecasting anomaly functions have been modified. Their new behavior is simpler (and more justifiable) than their old behavior.

They now return a value which expresses a standardized deviation from expectation. The anomaly flow now works like this:

* Gather data

* Estimate the data with a model

* Compute the difference between estimated and original data

* Normalize the differences by centering on 0, and then dividing by the sample standard deviation of the values.

Resulting values which are more than, say, 2, or less than, say, -2 could be considered anomalous.

TODO: anomaly fetches the data twice; we can eliminate this redundancy by adding some new function nodes for efficient sharing.

this is future work


function/forecast/anomaly_function.go
```diff
@@ -71,3 +77,82 @@ func FunctionPeriodicAnomalyMaker(name string, model function.MetricFunction) fu
...
+	meanDifference := 0.0
+	for _, difference := range differences {
+		meanDifference += difference
+	}
+	meanDifference /= float64(len(differences))
+
+	//
+	stddevDifference := 0.0
+	for _, difference := range differences {
+		stddevDifference += (difference - meanDifference) * (difference - meanDifference)
```
drcapulet: Should be able to just do `(difference - meanDifference) ** 2` here

drcapulet:
One comment, otherwise :100:

PR #241, opened 2015-11-27
Update Readme

Nathan-Fenner:
The readme has gotten a bit out of date.

It now has an update package structure (which accurately reflects the current package structure) and adds `gopeg` as a dependency.


syamp:
lgtm

PR #242, opened 2015-11-30
Rewrite query / Add aliasing syntax / deprecate transform.alias / fix response query syntax

Nathan-Fenner:
This PR adds a new syntax for aliasing queries:

You can now write

```
select (cpu.usage.percent / cpu.capacity) {CPU quota}
from -30m to now
```

Instead of

```
select cpu.usage.percent / cpu.capacity | transform.alias("CPU quota")
from -30m to now
```

The latter form would now be considered deprecated. (Although this PR maintains support for it)

Internally, this change simplifies the machinery required to define functions that manipulate `apiSeriesList`s. In particular, they no longer have a `Query` or `Name` field.

Instead, the `function.Expression` interface provides methods `QueryString()` and `Name()`.

It eliminates the need for every function to separately duplicate the code required to format them.

This change also therefore reduces the burden of tests to verify the correct formatting of their query/name strings. (Existing tests were incomplete- some functions were covered, while others were not).

Defining `transform.alias` as a function is neat, but not really sane. It was the only function that ever modified the name of a series list. So it was very, very special. It was also an overly long name considering that it was supposed to make queries easier to read (primarily in the UI).

### Internal Changes

This change was originally intended as a cleanup for internal manipulations of `api.SeriesList`s. `Name` and `Query` are no longer properties in the `api.SeriesList struct`. Instead, they're now defined as methods for the `function.Expression` interface.

The entirety of formatting is now handled by: `(f functionExpression) QueryString()`, so it doesn't need to be replicated anywhere else.

This also makes it unnecessary to ever check or validate individual function definitions' implementations of name assignment, since they can't control it. So we can write fewer tests, but have greater coverage.

### Syntax for Aliases

A new syntax has been added for aliases: curly-brace-enclosed strings that follow any expression. For example:

```
select
(cpu.usage.percent / cpu.usage.total) {cpu percent of quota}
from -30m to now
```

will alias the series list as "cpu percent of quota". 

It has the highest precedence of any operation. As a result, if you write `x {a} + y {b}` you'll get a result with the name `(a + b)`. It behaves as you'd probably expect when you use it with pipes: `x {a} | f` becomes `f(a)` and `x | f {a}` becomes just `a`.

The contents of the alias are literal, and, in particular, there's no way to escape the `}` character.

### What's become of `transform.alias`

In order to ensure backwards-compatibility, I special-cased the name `transform.alias` in the general function-printer implementation for `functionExpression`. This means that it will continue to work identically as before.

### Backwards-compatibility

As far as I can, from the user's perspective to existing queries, this only has one behavioral difference: if you accidentally use a series list where a string, duration, or number is expected, then the error can no longer tell you what the series list is called. As an example, if the user were to perform the incorrect query:

`transform.timeshift(disk, cpu)`

you used to get the error message:

`cannot convert serieslist[cpu] (type SeriesList) to type duration`

now you'll get the error message

`cannot convert serieslist (type SeriesList) to type duration`

which excludes the name of the invalid metric.

Fixing this error message (and others) to give more useful information will be a later project.

Error messages in general need improvement (especially syntax errors). It's probably worthwhile to think up a nicer error-handling system for expression evaluation.

### Fixing Syntax for Response Queries

The resulting responses now have correct syntax. This includes accurately surrounding the metric name in backticks (but only when necessary) and including any metric conditionals that may be present.


query/node.go
```diff
@@ -101,6 +144,58 @@ type functionExpression struct {
...
+		argumentQueries[i] = f.arguments[i].QueryString()
+	}
+	argumentString := strings.Join(argumentQueries, ", ")
+	groupString := ""
+	if len(f.groupBy) != 0 {
+		groupKeyword := "group by"
+		if f.groupByCollapses {
+			groupKeyword = "collapse by"
+		}
+		groupString = fmt.Sprintf(" %s %s", groupKeyword, strings.Join(f.groupBy, ", "))
```
drcapulet: Same thing here, probably need to properly escape each of `f.groupBy`

query/node.go
```diff
@@ -101,6 +144,58 @@ type functionExpression struct {
...
+	}
+	return fmt.Sprintf("%s(%s%s)", f.functionName, argumentString, groupString)
+}
+func (f functionExpression) Name() string {
+	// TODO: deprecate this behavior before it becomes permanent
+	if f.functionName == "transform.alias" && len(f.arguments) == 2 {
+		log.Printf("I'm an alias")
```
drcapulet: Not needed

query/node.go
```diff
@@ -101,6 +144,58 @@ type functionExpression struct {
...
+		if alias, ok := f.arguments[1].(*stringExpression); ok {
```
drcapulet: Where would this case fail? Also let's call `alias.QueryString()` and then we can have the contract that all `QueryString` responses should be properly escaped.

query/node.go
```diff
@@ -101,6 +144,58 @@ type functionExpression struct {
...
+	content    function.Expression
+	annotation string
+}
+
+func (ae annotationExpression) Evaluate(context *function.EvaluationContext) (function.Value, error) {
+	return ae.content.Evaluate(context)
+}
+
+func (ae annotationExpression) QueryString() string {
+	return fmt.Sprintf("%s {%s}", ae.content.QueryString(), ae.annotation)
```
drcapulet: Same here about the escaping

query/node.go
Nathan-Fenner: Unintentionally debugging that got left in - will remove

query/node.go
Nathan-Fenner: It fails if a user calls the function wrong. E.G,

`transform.alias(cpu, 2)` since 2 is not a string. Or, `transform.alias(cpu, cpu)`.

It's just defensive coding. An unchecked type assertion would raise a panic that kills the server if the user did that.

In this case, we actually don't want alias to be escaped- aliases are supposed to be human-readable, not surrounded by quotes. (That's why this behavior occurs in `Name` but not in `QueryString`).

Also, as I mentioned in the PR- I want to deprecate this functionality soon, I'm just not sure if there's any internal queries that still use `transform.alias`. So once they're all gone this whole `if` can be erased.

query/node.go
Nathan-Fenner: The contents of the annotation are always literal, so there's no need to escape.

This is because the annotation provides no way of escaping to begin with. In principle, the only illegal character is `}` and I didn't provide any way of escaping it in the parser. If we wanted to add one then this needs to change- but I don't think there's a significant benefit to allowing escaped characters for something that's supposed to be human-readable.

drcapulet: Ah yeah :+1:

query/node.go
drcapulet: Makes sense, we do use `alias` pretty extensively so we'll need to update our usages before we remove this wholesale.

query/node.go
Nathan-Fenner: done

drcapulet:
Nice job, LGTM :fire:

PR #243, opened 2015-12-01
Make the evaluation context no longer a pointer

Nathan-Fenner:
The evaluation context was passed around as a pointer, but never used like one. To this end, it is now passed as value everywhere instead of by pointer.

Hopefully, this makes reasoning about the context's behavior slightly easier, since modifications to the struct can no longer have any affect on other functions.

Performance
==========

Passing as a struct value is going to be more costly than passing as a pointer. Realistically, though, this will be dwarfed by the cost of evaluating even the simplest queries, even if it does have about a dozen fields.

Changes in Implementation
=====================

All the interface functions that took a `*function.EvaluationContext` now take a `function.EvaluationContext` instead.

The contents of the pointer were essentially never mutated, though it was occasionally copied in full for functions that modified timerange. To facilitate these, a method has been added called `WithTimerange(timerange)` to produces a new context which is identical, except for having the specified `timerange`.

Changes in Functionality
===================

This PR maintains previous functionality identically. It's just a refactoring of sorts.

@cchandler @drcapulet 

Taking it Further
============

Based on the observation that the majority of the fields of the `EvaluationContext` should be immutable, they can be made private. The #244 does this, as well as storing all of these private fields in a pointer (effectively eliminating the cost of the evaluation context passed by value, since almost all of the fields are passed by value).

#244 moves most of the fields of the `EvaluationContext` into a package-private variable called `internal`.


drcapulet:
Do any of these need to be / should be copied around as pointers on the `EvaluationContext`?

```
 	TimeseriesStorageAPI      api.TimeseriesStorageAPI // Backend to fetch data from
 	MetricMetadataAPI         api.MetricMetadataAPI    // Api to obtain metadata from
 	Timerange                 api.Timerange            // Timerange to fetch data from
 	SampleMethod              api.SampleMethod         // SampleMethod to use when up/downsampling to match the requested resolution
 	Predicate                 api.Predicate            // Predicate to apply to TagSets prior to fetching
 	FetchLimit                FetchCounter             // A limit on the number of fetches which may be performed
 	Cancellable               api.Cancellable
 	Registry                  Registry
```

Otherwise, LGTM.

Nathan-Fenner:
In short, everything should be fine. Almost all of the fields were already/still are interfaces or pointers, so almost no change:

*`TimeseriesStorageAPI` is an interface, so basically a pointer.
*`MetricMetadataAPI` is an interface, so basically a pointer.
*`Timerange` is treated as a value, so doesn't need to be a pointer.
*`SampleMethod` is treated as a value, so doesn't need to be a pointer.
*`Predicate` is an interface, so basically a pointer.
*`FetchLimit` is a struct, but it has a pointer inside, so passing around as value is okay.
*`Cancellable` is an interface, so basically a pointer.
*`Registry` is an interface, so basically a pointer.
*`Profiler` is an actual pointer
*`OptimizationConfiguration` is an actual pointer
*`EvaluationNotes` is an actual pointer

PR #244, opened 2015-12-04
EvaluationContext is not a pointer, and hides internals

Nathan-Fenner:
Please see #243 for the context of this change. This PR is a direct elaboration on #243.

The `EvaluationContext` struct becomes a proxy for the `TimeseriesStorageAPI` and the `MetricMetadataStorageAPI`, and the `Registry`, as well as other features.

This makes the `EvaluationContext` itself more complex, but now it does everything that could be desired of it.

It's not clear whether this helps or hurts the codebase complexity.


Nathan-Fenner:
Going to close this for now- the benefits aren't significant enough for the size of the changes.

PR #245, opened 2015-12-07
Add messages describing the source of type conversion errors

Nathan-Fenner:
When a type conversion error occurs, now the query that produced the value with the wrong type is printed, instead of the value itself.

This mainly serves to fix the issue associated with removing names from SeriesLists- now when you accidentally use a serieslist as a (say) duration, the invalid argument will be identified by name in the error.

@drcapulet


drcapulet:
:+1:

PR #246, opened 2015-12-10
Remove unnecesarry escaping for names with dots

Nathan-Fenner:
A recent change which allowed the query labels to correctly escape their contents was overzealous. In particulae, `foo.bar` would always be escaped because of the `.`, even though this is a legal identifier.

This PR updates the identifier regex used in escaping to reflect this, and updates the tests to correct this oversight.

@drcapulet


drcapulet:
:+1:

PR #247, opened 2015-12-15
Cleanup & Changes to Optimization

Nathan-Fenner:
Originally, I had intended to do just a little bit of cleanup.

The main changes are

* Minor miscellaneous style changes to improve consistency
* The `optimize` package is now less intrusive: it modifies the execution context once, up front, instead of being referred to in the body of various query functions
* Make Blueflood's concurrent-fetching implementation available as a generic optimization

In particular, the `optimize` package now wraps the `MetricMetadataAPI` in a new "caching metadata API" that adds a cache for the results of the `GetAllTags()`. The cache is 2 hours by default, the same as the previous implementation.

It also adds the `NewParallelTimeseriesStorageAPI` which embeds a `TimeseriesStorageAPI`, but wraps its calls to `FetchMultipleTimeseries` so that they're executed concurrently.

The Blueflood API implementation is already fairly complicated (since it tries to do things like intelligent resolution selection), and made much worse by the logic for parallel fetches.

This PR removes this logic from the Blueflood API implementation and puts it into the `optimize` package. Thus the Blueflood API, without optimization enabled, will run sequentially (i.e. very slowly). Thus performance is now dependent on having this optimization enabled.

TODO: Test the `NewParallelTimeseriesStorageAPI`

@cchandler @alexc


api/timeseries_storage_api.go
```diff
@@ -31,7 +31,6 @@ type FetchTimeseriesRequest struct {
 	Metric                TaggedMetric // metric to fetch.
 	SampleMethod          SampleMethod // up/downsampling behavior.
 	Timerange             Timerange    // time range to fetch data from.
-	MetricMetadata        MetricMetadataAPI
```
Nathan-Fenner: The API data wasn't actually being used, so it has been removed. This had very little effect on existing code.

function/expression.go
```diff
@@ -19,18 +18,17 @@ import (
...
-	TimeseriesStorageAPI      api.TimeseriesStorageAPI // Backend to fetch data from
-	MetricMetadataAPI         api.MetricMetadataAPI    // Api to obtain metadata from
-	Timerange                 api.Timerange            // Timerange to fetch data from
-	SampleMethod              api.SampleMethod         // SampleMethod to use when up/downsampling to match the requested resolution
-	Predicate                 api.Predicate            // Predicate to apply to TagSets prior to fetching
-	FetchLimit                FetchCounter             // A limit on the number of fetches which may be performed
-	Cancellable               api.Cancellable
-	Registry                  Registry
-	Profiler                  *inspect.Profiler // A profiler pointer
-	OptimizationConfiguration *optimize.OptimizationConfiguration
```
Nathan-Fenner: The real change is the removal of the `OptimizationConfiguration` field (the others just have new indentation).

Now, OptimizationConfigurations manipulate the `Context` directly, so there's no need for it to it to be stored or accessed during evaluation

main/common/common.go
```diff
@@ -34,7 +34,6 @@ import (
 var (
 	// ConfigFile is the location of the rule YAML file.
 	ConfigFile = flag.String("config-file", "", "Location of YAML config file")
-	Verbose    = flag.Bool("verbose", false, "Set to true to enable logging")
```
Nathan-Fenner: The verbose flag was not used.

main/ui/ui.go
```diff
@@ -84,21 +84,20 @@ func main() {
...
-		OptimizationConfiguration: optimizer,
-		UserSpecifiableConfig:     userConfig,
+	optimizer := optimize.NewOptimizationConfiguration()
+	executionContext := optimizer.OptimizeExecutionContext(query.ExecutionContext{
+		MetricMetadataAPI:     apiInstance,
+		TimeseriesStorageAPI:  blueflood,
+		FetchLimit:            1500,
+		SlotLimit:             5000,
+		Registry:              registry.Default(),
+		UserSpecifiableConfig: userConfig,
```
Nathan-Fenner: This is where the fundamental difference in how optimization works occurs:

Now, a preliminary `ExectionContext` is passed to `optimizer.OptimizeExecutionContext`, which returns a magically-optimized version (it does this by replacing the APIs with faster versions).

PR #248, opened 2016-01-09
Define Config Structs in Consistent Packages

Nathan-Fenner:
Now, Cassandra's config struct is defined in the `cassandra` package, and the UI's config struct is defined in the `ui` package.

**THIS IS A BREAKING CHANGE**: configs will need to be slightly modified.

See an [example config file](https://github.com/square/metrics/blob/nathan-fenner-common-config-remove/demo/example-config.yaml) to understand how it's laid out.

The main change is `api => cassandra`, the removal of the `config` field from `ui`, and pulling `conversion_rules` out to the top, instead of making it a member of `api` (which actually was the only part that *didn't* use it).

 @achow @cchandler @drcapulet


demo/example-config.yaml
```diff
@@ -1,4 +1,6 @@
 
+conversion_rules_path: demo/conversion_rules  # the directory for the conversion rules
```
drcapulet: Seems like this might be better under the blueflood config section since it's only used there?

drcapulet:
One comment, otherwise LGTM.

Nathan-Fenner:
I'm going to go forward with this change as-is, waiting for #251 to address the above comment by removing conversion rules from Blueflood

PR #250, opened 2016-01-12
Add 'tolerate' feature to autocom

Nathan-Fenner:
If a user has something like
```
    `cpu.percent`^
```
where the `^` is the cursor, and press enter they get

```
    cpu.percent^
```

instead of what they (probably) expect,

```
    `cpu.percent`
    ^
```

This PR fixes this issue.

In addition, it offers a slight performance improvement when there's a large number of options, by calling the autocompletion function about 3x less.

@drcapulet 

(Tested locally)

### DIFF
--- a/main/static/autocom.js
+++ b/main/static/autocom.js
@@ -37,19 +37,23 @@ window.Autocom = (function(){
 	var colorCom = new Autocom(input);
 
 	// Assign the options for our Autocom
-	colorCom.options = ["red", "orange", "yellow", "yellow-green", "green", "cyan", "blue", "purple"];
-	
+	colorCom.options = ["red", "orange", "yellow", "yellow-green", "green", "cyan", "blue", "purple", "gray"];
+
 	// Words are made of lowercase letters and hyphens, but can't start with a hyphen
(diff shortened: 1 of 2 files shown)


main/static/autocom.js
```diff
@@ -286,11 +287,9 @@ function generateTooltipContents(tooltip, words, index, selectedCallback) {
 // Inserts `word` at the location specified by `at` inside of `input`.
 function insertWord(input, at, word, supressCallback) {
 	input.value = input.value.substring(0, at.from) + word + input.value.substring(at.to);
-	setTimeout(function() {
```
drcapulet: Any reason you moved this out of the `setTimeout`?

Nathan-Fenner: Found out that it wasn't necessary.

I can't remember why it was added originally, so I assume it was because the browser was preventing the write for some reason or another, or that it was causing an infinite loop. Either way, doesn't need t be there anymore.

You'd notice it a little, because after pressing `enter` to autocomplete, you'd see the cursor jump a little. No point in doing that, so got rid of it.

drcapulet:
:+1:

PR #252, opened 2016-01-19
Add more profiling

Nathan-Fenner:
Adds profiling for parsing and a few other measurements. Minor refactor to make this possible.

@drcapulet

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -150,6 +150,8 @@ func (h tokenHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reques
 }
 
 func (q queryHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	profiler := inspect.New()
+	serveHTTPTimerCompletion := profiler.Record("Server HTTP")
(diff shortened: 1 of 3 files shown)


ui/ui.go
```diff
@@ -179,6 +186,7 @@ func (q queryHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reques
 		Metadata: result.Metadata,
 		Name:     cmd.Name(),
 	}
+	serveHTTPTimerCompletion()
```
drcapulet: Should this be after the `OnQuery` hook?

drcapulet:
:+1:

ui/ui.go
Nathan-Fenner: It can't go any lower, since we want to include the profiler's results in `response` (assigned just below this).

`convertProfile` needs the profiler to have all its timers finished, and the profile data needs to be ready for being put in the `response` to be written (again, before the hook).

drcapulet: Ah, yeah good point, too bad be can't look into the future.

PR #253, opened 2016-01-20
Add context-awareness to autocompletion

Nathan-Fenner:
Namely, the use of `where` or `from` or `to` or `resolution` or `sample` will cause it to stop suggesting metric names or functions to fill in.

It's slightly brittle, since it's just looking for `\swhere\s` (which can appear in backquoted metrics, for example) or can fail to be detected because of something like `)where`...

In either case, losing (or keeping) autocomplete support is not the end of the world. Fixing this issue in all cases would require a complete parser for the query grammar, which is currently outside the scope of the client-side JS.

(Tested locally)

@drcapulet


drcapulet:
:+1:

PR #255, opened 2016-02-16
Reduce Strength of Caching on MQE

Nathan-Fenner:
Caching has some negative consequences for metrics when new tag-sets are added, making it look like it takes much longer to propagate into MQE than it really should.

However, the latency from performing the tag-set query for each metric every time can be quite large for metrics with lots of time-series.

## `CachedMetricMetadataAPI`

This PR introduces the `CachedMetricMetadataAPI` wrapper, which holds onto an underlying `MetricMetadataAPI` which it uses to perform queries. This cache system supersedes the `optimize` package, eliminating the new for MQE code to be cognizant or aware of these caching- its presents or absence is completely transparent to the system.

In order to minimize latency, the `CachedMetricMetadataAPI` serves responses from its cache whenever possible (unless they are absent or out-of-date). However, even when they are present in the cache, it still performs a request (on a separate goroutine) to the underlying `MetricMetadataAPI` in order to keep the cache as fresh as possible.

This means individual queries will retain low-latency behavior, while the cache can be forced to update simply by waiting for the request to go through.

Lastly, in order to minimize the potential for excess strain on the underlying API due to an excessive number of requests, the `CachedMetricMetadataAPI` has a `RequestLimit` which determines the maximum number of in-flight requests at a time. If the request limit is reached, then background-updates are dropped to try to reduce request pressure.

## No more `RemoveMetric`

It wasn't used anywhere in the codebase, apart from `Cassandra` tests. Hence it has been removed to reduce the need for continuing to maintain it.

@drcapulet


metric_metadata/cached_metadata/cached_test.go
```diff
@@ -0,0 +1,146 @@
...
+	{
+		underlying.proceed <- struct{}{} // give it permission to go
+		answer, err := cached.GetAllTags("metric_two", api.MetricMetadataAPIContext{})
+		a.CheckError(err)
+		a.Eq(answer[0]["foo"], "two")
+		a.Contextf("underlying count").Eq(underlying.count, 2)
+		<-underlying.finished
+	}
+
+	// Ask for metric_one again, but do not give permission to go.
```
drcapulet: Maybe add a comment here that its because we want to make sure it pulls from the cache and not the underlying call?

metric_metadata/cassandra/cassandra_api_test.go
```diff
@@ -94,11 +94,12 @@ func Test_MetricName_GetTagSet_API(t *testing.T) {
 				api.ParseTagSet(c.tagString),
 			}, context))
 		} else {
-			a.CheckError(cassandra.RemoveMetric(api.TaggedMetric{
 
+			a.CheckError(cassandra.db.session.Query(
```
drcapulet: nit: maybe make this a helper only available in test?

Nathan-Fenner: yeah- added a function for it

metric_metadata/cached_metadata/cached_test.go
Nathan-Fenner: added a note

drcapulet:
LGTM :fire:

PR #256, opened 2016-02-19
Filter Threshold & Filter Count Functions

Nathan-Fenner:
# Threshold Filtering

You can now say:

```
select
cpu | filter.mean_above(25, 1hr)
```

to select all CPU timeseries whose mean in the last hour is above 25. Or, just

```
select
cpu | filter.mean_above(25)
```

to only use the currently queried time interval.

# This is a breaking change!

Instead of having both functions

```
filter.highest_max(series, count)
filter.recent_highest_max( series, count, duration )
```

this PR replaces them with the new definition for `filter.highest_max`:

```
filter.highest_max( series, count [, duration] )
```

which now acts as both versions.

@drcapulet @syamp

### DIFF
--- a/api/timerange.go
+++ b/api/timerange.go
@@ -59,7 +59,7 @@ func (tr Timerange) EndMillis() int64 {
 // End returns the time.Time value corresponding to the end of the timerange (inclusive).
 func (tr Timerange) End() time.Time {
 	seconds := tr.end / 1000
-	nanoseconds := (tr.start % 1000) * 1000000
+	nanoseconds := (tr.end % 1000) * 1000000
 	return time.Unix(seconds, nanoseconds)
 }
 
--- a/api/timerange.go
+++ b/api/timerange.go
@@ -174,3 +174,17 @@ func (tr Timerange) ExtendBefore(length time.Duration) Timerange {
 func (tr Timerange) Slots() int {
 	return int((tr.end-tr.start)/tr.resolution) + 1
 }
+
+// TimeOfIndex returns the point in time corresponding to a (possibly out-of-range)
+func (tr Timerange) TimeOfIndex(point int) time.Time {
+	return tr.Start().Add(time.Duration(point) * tr.Resolution())
+}
+
+// IndexOfTime returns the index of the point in time (possibly out-of-range).
+// 0 corresponds to [tr.Start(), tr.Start().Add(tr.Resolution()))
+func (tr Timerange) IndexOfTime(point time.Time) int {
+	if tr.Resolution() == 0 {
+		return 0
+	}
+	return int(point.Sub(tr.Start()) / tr.Resolution())
+}
--- a/api/timerange.go
+++ b/api/timerange.go
@@ -176,12 +176,13 @@ func (tr Timerange) Slots() int {
 }
 
 // TimeOfIndex returns the point in time corresponding to a (possibly out-of-range)
+// index. N corresponds to tr.Start + N*tr.Resolution.
 func (tr Timerange) TimeOfIndex(point int) time.Time {
 	return tr.Start().Add(time.Duration(point) * tr.Resolution())
 }
 
 // IndexOfTime returns the index of the point in time (possibly out-of-range).
-// 0 corresponds to [tr.Start(), tr.Start().Add(tr.Resolution()))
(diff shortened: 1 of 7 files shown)


api/timerange.go
```diff
@@ -53,8 +53,8 @@ func (tr Timerange) End() int64 {
 }
 
 func (tr Timerange) EndTime() time.Time {
-	seconds := tr.End() / 1000
-	milliseconds := tr.Start() % 1000
+	seconds := tr.end / 1000
+	milliseconds := tr.end % 1000
```
Nathan-Fenner: This line was actually wrong before- it caused a test to suddenly start failing when I made a (seemingly benign) change.

api/timerange.go
```diff
@@ -175,3 +174,17 @@ func (tr Timerange) ExtendBefore(length time.Duration) Timerange {
 func (tr Timerange) Slots() int {
 	return int((tr.end-tr.start)/tr.resolution) + 1
 }
+
+// TimeOfIndex returns the point in time corresponding to a (possibly out-of-range)
```
drcapulet: Unfinished thought?

Nathan-Fenner: Oops- I'll finish that

api/timerange.go
```diff
@@ -175,3 +174,18 @@ func (tr Timerange) ExtendBefore(length time.Duration) Timerange {
...
+// index. N corresponds to tr.Start + N*tr.Resolution.
```
drcapulet: By N I assume you mean `point`? Might be better to rename it `index int` and match the comment + signature.

function/filter/filter.go
```diff
@@ -35,6 +35,9 @@ func (list filterList) Less(i, j int) bool {
 	if math.IsNaN(list.value[j]) && !math.IsNaN(list.value[i]) {
 		return true
 	}
+	if math.IsNaN(list.value[j]) {
```
drcapulet: Not exactly sure how we're defining, but isn't the `if` block above now useless?

function/registry/registry.go
```diff
@@ -155,42 +159,55 @@ func NewFilter(name string, summary func([]float64) float64, ascending bool) fun
...
+					return nil, err
+				}
+			}
 			// Round to the nearest integer.
 			count := int(countFloat + 0.5)
 			if count < 0 {
 				return nil, fmt.Errorf("expected positive count but got %d", count)
 			}
-			result := filter.FilterBy(list, count, summary, ascending)
+			if recentDuration < 0 {
```
drcapulet: Is there ever a case where `0` is a valid value?

testing_support/mocks/combo_api.go
```diff
@@ -0,0 +1,135 @@
...
+}
+func (fapi FakeComboAPI) CheckHealthy() error {
+	return nil
+}
+
+var _ metadata.MetricAPI = FakeComboAPI{}
+
+func (fapi FakeComboAPI) ChooseResolution(requested api.Timerange, smallestResolution time.Duration) time.Duration {
+	if fapi.timerange.Resolution() < smallestResolution {
+		//	panic("ChooseResolution is too coarse for FakeComboAPI instance.")
```
drcapulet: Mean to leave this commented out?

function/filter/filter.go
Nathan-Fenner: It looks like one of these is actually backwards. Good catch!

api/timerange.go
Nathan-Fenner: fixed

function/filter/filter.go
Nathan-Fenner: fixed

function/registry/registry.go
Nathan-Fenner: I think it's endpoint-inclusive, so this should mean "filter using only the very last point in the series."

I don't think it's very obvious but it should work.

testing_support/mocks/combo_api.go
Nathan-Fenner: fixed

function/filter/filter.go
drcapulet: Is it easy to add a test that would have caught this?

Nathan-Fenner: I've added a test for it.

drcapulet:
LGTM

PR #262, opened 2016-03-11
Fix unintentionally creating lots of goroutines with cached metadata

Nathan-Fenner:
A huge number of goroutines could be spawned to perform background requests. (They wouldn't actually "leak" because they'd eventually either be handled or dropped, but this is still a waste of resources).

Now the `CachedMetricMetadataAPI` no longer spawns goroutines. Instead, the consumer of the API instance has to invoke the `.PerformBackgroundRequest() error` method explicitly. This is a blocking method, so the caller should do something like this:

```

cachedAPI := cached_metadata.NewCachedMetricMetadataAPI(underlyingAPI, cached_metadata.Config {
    TimeToLive: time.Hour * 2,
    RequestLimit: 500,
})

for i := 0; i < 10; i++ {
    go func() {
        for {
            err := cachedAPI.PerformBackgroundRequest(
                api.MetricMetadataContext{Profile: profiler}
            )
            if err != nil {
                log.Errorf("Error performing background request: %s", err.Error())
            }
        }
    }() 
}
```

Background requests are served on a first-come, first-served basis. If there are too many, then the most recent ones will be dropped (they are stored in a channel). In particular, this won't cause calls to `cached.GetAllTags(metric, context)` to block any longer than normal.

@syamp @drcapulet

### DIFF
--- a/main/ui/ui.go
+++ b/main/ui/ui.go
@@ -84,6 +84,16 @@ func main() {
 		TimeToLive:   time.Minute * 5, // Cache items invalidated after 5 minutes.
 		RequestLimit: 500,
 	})
+	for i := 0; i < 10; i++ {
+		go func() {
+			for {
+				err := optimizedMetadataAPI.PerformBackgroundRequest(api.MetricMetadataAPIContext{})
+				if err != nil {
+					log.Errorf("Error performing background cache-update: %s", err.Error())
+				}
+			}
+		}()
+	}
 
 	//Defaults
 	userConfig := api.UserSpecifiableConfig{
--- a/metric_metadata/cached_metadata/cached.go
+++ b/metric_metadata/cached_metadata/cached.go
@@ -15,7 +15,6 @@
 package cached_metadata
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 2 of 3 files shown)


syamp:
I'd like us to review using groupcache in favor of internal MQE caching.

main/ui/ui.go
```diff
@@ -84,6 +84,16 @@ func main() {
 		TimeToLive:   time.Minute * 5, // Cache items invalidated after 5 minutes.
 		RequestLimit: 500,
 	})
+	for i := 0; i < 10; i++ {
```
syamp: why 10

Nathan-Fenner: It's arbitrary, I guess. Previously there was an explosion of goroutines doing background fetches, but you probably only need a handful.

You can handle 1 query per goroutine, so at "X" QPS (counting number of metrics named in MQE queries), the latency has to be at most 10 / X seconds to keep up.

So with 100 QPS, each goroutine has to handle 10 requests per second, capping you at 100ms maximum latency to keep up (or some background cache updates will be dropped).

This code isn't actually used in the Square internal repo, so this being non-configurable at the moment is not such a big deal.

Nathan-Fenner:
Groupcached doesn't allow values to change:

> does not support versioned values. If key "foo" is value "bar", key "foo" must always be "bar". **There are neither cache expiration times, nor explicit cache evictions**. Thus there is also no CAS, nor Increment/Decrement. This also means that groupcache....

We can't use it to solve this problem, which is that the caching is frustrating because it updates infrequently.

main/ui/ui.go
Nathan-Fenner: In addition, any client using MQE as a library (which is probably a lot more likely, in order to get it to integrate with other systems better) wouldn't be using this either.

drcapulet:
:+1:

PR #269, opened 2016-03-25
Add CheckHealthy methods for metadata.MetricAPI and timeseries.StorageAPI

drcapulet:
Although the server won't boot up without being able to talk to Cassandra, if if goes away it'd be nice to be able to check the status of the underlying connections in the MetricMetadataAPI.

R: @Nathan-Fenner || @syamp

### DIFF
--- a/metric_metadata/api.go
+++ b/metric_metadata/api.go
@@ -39,4 +39,6 @@ type MetricAPI interface {
 	// GetMetricsForTag takes a tag key-value pair and returnsthe list of all the
 	// MetricKeys associated with them.
 	GetMetricsForTag(tagKey, tagValue string, context Context) ([]api.MetricKey, error)
+	// CheckHealthy checks if this MetricAPI is healthy, returning a possible error
+	CheckHealthy() error
 }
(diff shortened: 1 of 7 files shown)


Nathan-Fenner:
It seems like a good idea.

Would it be better to return an `error` so that whoever checks it can (at least hopefully) get some indication of what went wrong?

drcapulet:
Yeah, originally I was thinking that there wouldn't be a case where the health check would succeed but other requests fail. However, if we add other backends for the metadata API they could have varying failure cases.

What do you think about the name CheckHealth and have it return an error?

Nathan-Fenner:
I think that makes sense. `CheckHealth() error` seems like a good interface for it, and the name seems good.

Nathan-Fenner:
It actually would make sense to do the same thing for the `TimeseriesStorageAPI`.

drcapulet:
Good idea, I'll add it there as well - for Blueflood the only way to check is with a  `GET /v2.0` so that'll have to do.

PR #274, opened 2016-04-07
Rework CachedMetricMetadataAPI

drcapulet:
* Adds in concept of Stale (by default disabled) whereupon if the cache entry is older than that a background fetch request will be enqueued.
* Any request for a metric key where a fetch is already in progress and the cache item has expired will wait for that fetch to come back (instead of issuing it's own synchronous call).
* Added some more logging.

R: @Nathan-Fenner || @DanielNi || @syamp || @cuzelac


metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,114 @@
...
+// addBackgroundGetAllTagsRequest adds a job to update the lag list for the given
+// metric. Requires the client hold the lock for the item in the cache
+func (c *CachedMetricMetadataAPI) addBackgroundGetAllTagsRequest(metricKey api.MetricKey) {
+	c.queueMutex.Lock()
+	defer c.queueMutex.Unlock()
 
-func (c *CachedMetricMetadataAPI) addBackgroundGetAllTagsRequest(metric api.MetricKey) {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
 	if len(c.backgroundQueue) < cap(c.backgroundQueue) {
```
Nathan-Fenner: I realize now that it's more idiomatic for this to say 
```
if len(c.backgroundQueue) >= cap(c.backgroundQueue) {
     log.Warningf("Unable to enqueue a background GetAllTags lookup for %s due to a full queue", metricKey)
     return
}
```

to avoid the additional nesting

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,114 @@
...
-		// Add this function to the background queue.
+		item, ok := c.getAllTagsCache[metricKey]
```
Nathan-Fenner: Need to lock the mutex to access map: reads are not thread-safe in the present of writes.

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,114 @@
...
+		if !ok {
+			log.Errorf("Asked to perform a background GetAllTags lookup for %s but missing entry", metricKey)
+			return
+		}
+
+		if item.enqueued {
```
Nathan-Fenner: Don't we need to lock the `item`'s mutex before reading/writing to it?

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,114 @@
...
+		}
+
+		log.Infof("Enqueuing a background GetAllTags lookup for %s", metricKey)
+		item.enqueued = true
+
 		c.backgroundQueue <- func(context metadata.Context) error {
+			log.Infof("Executing the background GetAllTags lookup for %s", metricKey)
+			defer log.Infof("Finished the background GetAllTags lookup for %s", metricKey)
+
+			item.Lock()
```
Nathan-Fenner: Never unlocked?

metric_metadata/cached/cached.go
```diff
@@ -111,37 +157,45 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
-	}
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-	if c.getAllTagsCache[metricKey].Expiry.Before(value.Expiry) {
-		c.getAllTagsCache[metricKey] = value
+// fetchAndUpdateCachedTagSet updates the in-memory cache (asusming the update)
+// is newer than what is in the cache. The caller MUST have the lock on the item
+// in the cache already and will release that lock before returning
+func (c *CachedMetricMetadataAPI) fetchAndUpdateCachedTagSet(metricKey api.MetricKey, context metadata.Context) ([]api.TagSet, error) {
+	item, ok := c.getAllTagsCache[metricKey]
```
Nathan-Fenner: need to log cache mutex

Nathan-Fenner:
I've made some notes where I think there are some race conditions or cleanup to be done. I haven't finished over it, but I think it might be possible to make this a bit simpler by changing the structure to be:

`CacheData map[MetricKey]Data`

where `Data` stores the fields and the time that they were updated, and nothing else.

`CacheStatus map[MetricKey]Status`

`CacheArrival map[MetricKey] <-chan struct{}`

where `Status` is one of `Absent` (zero-value), `Stored` (for metrics with `CacheData`), `Enqueued` (for metrics that may or may not have CacheData but are awaiting a request to update them) and `InFlight` (for metrics that may or may not have a stored value but have an in-flight request).

To synchronize waits for `InFlight` requests, receive on a member of the `CacheArrival` map (which is protected by the same mutex as `CacheStatus`).

When a request comes in, we check the cache:

* If fresh, return from `CacheData` and do nothing
* If not expired but not fresh, return from `CacheData` but update `CacheStatus` to `Enqueued` (unless it's already `InFlight`, in which case, do nothing).
* If expired, update `CacheStatus` to `InFlight` and run it yourself and return the result.

In particular, there should be functions:

`PromoteToQueue` (fast) : locks the maps, checks if it's not `InFlight`, and then sets it to `Enqueued`.

`PromoteToFlight` (blocks until done with request): locks the maps, checks if it's not `InFlight`:

* If it's not `InFlight`, set it to be `InFlight`, make a channel, set it for `CacheArrival` on that key, then unlock the mutex. Launch the request and when you're done, store it in `CacheData`, and close the synchronizing `chan`. Lastly, as a convenience, return the value.
* If it already was `InFlight`, get the `chan` instance from `CacheArrival` for this metric, unlock the mutex, and wait on the channel. Then lock, read `CacheData`, unlock, and return the result.

metric_metadata/cached/cached.go
drcapulet: This method requires the caller to already have the lock on the given item.

metric_metadata/cached/cached.go
drcapulet: We already have the lock on the item (see the method definition).

metric_metadata/cached/cached.go
drcapulet: Handled by `fetchAndUpdateCachedTagSet`

metric_metadata/cached/cached.go
drcapulet: Already locked, this method requires the caller to have the lock.

drcapulet:
AFAICT there aren't any race conditions, certain methods just require the lock to be in a specific state.

This is ideally focused around the least number of "global" locks - hence the only ones are `getAllTagsCacheMutex` for protecting insertion to `getAllTagsCache` and `queueMutex` for protecting the queue. Everything else simply requires the item's lock.

As a side note, we can't use channels here because we can have n threads all waiting on the same `GetAllTags` request.

Any specific reason you prefer the poll method (like you described, where you have a process polling for which metrics need an update) over a push method (by pushing jobs onto a queue)? Also, any specific reason for splitting things out into three maps instead of just one?

Nathan-Fenner:
I see some things that are still race conditions:

```
func (c *CachedMetricMetadataAPI) GetAllTags(metricKey api.MetricKey, context metadata.Context) ([]api.TagSet, error) {
 	defer context.Profiler.Record("CachedMetricMetadataAPI_GetAllTags")()

 	// Get the cached result for this metric.
	item, ok := c.getAllTagsCache[metricKey]
```

Map access isn't threadsafe- need to acquire a lock before reading (it's not atomic, so you can get garbage data or a panic).

>As a side note, we can't use channels here because we can have n threads all waiting on the same `GetAllTags` request.

I was suggesting waiting for a to channel close (which lots of things can listen to) instead of actually using a received value.

> Any specific reason you prefer the poll method (like you described, where you have a process polling for which metrics need an update) over a push method (by pushing jobs onto a queue)?

In this case, I think it just makes things simpler, because it means the map that holds item status functions as a queue (so you don't need to keep them in sync- marking something as "needs to be updated" is equivalent to adding it to the "queue").

>Also, any specific reason for splitting things out into three maps instead of just one?

It's probably better to combine them together. You could get away with only using one lock (though in principle, two locks, one for `CacheData` and one for the other two, would suffice).

drcapulet:
> Map access isn't threadsafe- need to acquire a lock before reading (it's not atomic, so you can get garbage data or a panic).

Ah, nice catch - I've updated that.

> I was suggesting waiting for a to channel close (which lots of things can listen to) instead of actually using a received value.

Gotcha, I do like the explicitness of a `sync.WaitGroup` though.

> In this case, I think it just makes things simpler, because it means the map that holds item status functions as a queue (so you don't need to keep them in sync- marking something as "needs to be updated" is equivalent to adding it to the "queue").

<div></div>

> It's probably better to combine them together. You could get away with only using one lock (though in principle, two locks, one for `CacheData` and one for the other two, would suffice).

So the downside of that approach is that to check the status of any individual item you need to lock the entire "table", whereas by only locking on the original write to make sure the item exists and then all other operations are locked on a per-item basis. It does improve the queue approach (instead of having a channel you pull functions from), but we're doing many people reads than updates.

metric_metadata/cached/cached.go
```diff
@@ -111,37 +160,47 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
+	item.Unlock()
+
+	startTime := c.clock.Now()
 	tagsets, err := c.metricMetadataAPI.GetAllTags(metricKey, context)
 	if err != nil {
+		item.Lock()
+		item.fetchError = err
+		item.wg.Done()
+		item.inflight = false
+		item.Unlock()
```
Nathan-Fenner: More idiomatic to `defer` the Unlock right before we lock it just above this.

metric_metadata/cached/cached.go
```diff
@@ -111,37 +160,47 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
+	newExpiry := startTime.Add(c.timeToLive)
+	if item.Expiry.Before(newExpiry) {
+		item.TagSets = tagsets
+		item.Expiry = newExpiry
+		item.Stale = startTime.Add(c.freshness)
+	}
+
+	item.wg.Done()
+	item.inflight = false
+	item.Unlock()
```
Nathan-Fenner: more idiomatic to `defer` the Unlock right after we lock it.

metric_metadata/cached/cached.go
```diff
@@ -111,37 +160,47 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
-		Expiry:  startTime.Add(c.timeToLive),
-	}
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-	if c.getAllTagsCache[metricKey].Expiry.Before(value.Expiry) {
-		c.getAllTagsCache[metricKey] = value
+// fetchAndUpdateCachedTagSet updates the in-memory cache (asusming the update)
+// is newer than what is in the cache. The caller MUST have the lock on the item
+// in the cache already and will release that lock before returning
+func (c *CachedMetricMetadataAPI) fetchAndUpdateCachedTagSet(metricKey api.MetricKey, context metadata.Context) ([]api.TagSet, error) {
```
Nathan-Fenner: I think this should take the `item` as an argument (as a `*CachedTagSetList`) instead of looking it up in the map. In both places that it's called, the `item` pointer has already been figured out.

Note: as-is, there's a race condition in this function: it accesses `getAllTagsCache` without a lock (neither caller acquires one for it). By avoiding the map access completely (by taking `item` as a parameter), we don't have to worry about this.

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,117 @@
...
-		}
+	// No need to lock on getAllTagsCache since we already should hold the lock for
+	// the given item, so it should have already been written in a thread-safe way
+	item, ok := c.getAllTagsCache[metricKey]
+	if !ok {
+		log.Errorf("Asked to perform a background GetAllTags lookup for %s but missing entry", metricKey)
+		return
+	}
+
+	if item.enqueued {
```
Nathan-Fenner: We should acquire the mutex on `item` here, instead of below, BEFORE we read `enqueued` and `inflight`, instead of after.

However, this makes unlocking it take a bit more work. I'd suggest that `fetchAndUpdateCachedTagSet` not return with the item's lock in an unlocked state (it's still allowed to unlock/lock again, just don't unlock when returning). Then we can just trivially `defer` the unlock here, (and in the other place that it's called).

To clarify: `fetchAndUpdateCachedTagSet` can still do its unlocking/locking, it just leaves things in a locked state instead of an unlocked one.

metric_metadata/cached/cached.go
drcapulet: Yeah realized this is duplicated, so just moved it right after the `GetAllTags` call.

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,117 @@
...
+
+	if item.inflight {
+		log.Infof("Unable to perform a background GetAllTags lookup for %s as one is already in flight", metricKey)
+		return
+	}
+
+	log.Infof("Enqueuing a background GetAllTags lookup for %s", metricKey)
+	item.enqueued = true
+
+	c.backgroundQueue <- func(context metadata.Context) error {
```
drcapulet: It's been moved up to lines 92-95.

metric_metadata/cached/cached.go
drcapulet: So this method requires the caller to already have the lock, I'll change them both to leave it in a locked state.

metric_metadata/cached/cached.go
drcapulet: I can definitely pass it as an argument, though there shouldn't be a race condition since we already have the lock on the given item and `getAllTagsCache` is insert only so we know the pointer hasn't changed.

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,118 @@
...
-			tagsets, err := c.metricMetadataAPI.GetAllTags(metric, context)
-			if err != nil {
-				return err
-			}
-			c.replaceCachedTagSet(metric, tagsets, startTime)
-			return nil
-		}
+	// No need to lock on getAllTagsCache since we already should hold the lock for
+	// the given item, so it should have already been written in a thread-safe way
+	item, ok := c.getAllTagsCache[metricKey]
```
Nathan-Fenner: There's still nothing protecting this read:

Just because we have a lock on `item` doesn't mean that someone isn't modifying a different metricKey in the cache: this causes a race condition.

We either need to acquire the `getAllTagsCacheMutex` before this read, or we need the caller to do it.

Since in all cases that this function is called, the caller already retrieved `item` from the map themselves a few lines, it seems simpler to me to just ask for `item` as a parameter and avoid the map lookup completely.

metric_metadata/cached/cached.go
```diff
@@ -15,68 +15,115 @@
...
 
 // CachedMetricMetadataAPI caches some of the metadata associated with the API to reduce latency.
 // However, it does not reduce total QPS: whenever it reads from the cache, it performs an update
 // in the background by launching a new goroutine.
 type CachedMetricMetadataAPI struct {
-	metricMetadataAPI metadata.MetricAPI                 // The internal MetricMetadataAPI that performs the actual queries.
-	getAllTagsCache   map[api.MetricKey]CachedTagSetList // The cache of "tags from metric name"
-	timeToLive        time.Duration                      // Time for cache entries to survive.
-	mutex             sync.Mutex                         // Synchronizing mutex
+	metricMetadataAPI metadata.MetricAPI // The internal MetricMetadataAPI that performs the actual queries.
```
cuzelac: fix comment: use current type

metric_metadata/cached/cached.go
```diff
@@ -111,70 +158,121 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
-	return c.getAllTagsCache[metricKey]
-}
-
-// setCachedTagSet is a thread-safe way to assign cached data (protected by a mutex).
-// if the given value is less recent than the stored value, nothing happens.
-func (c *CachedMetricMetadataAPI) replaceCachedTagSet(metricKey api.MetricKey, tagsets []api.TagSet, startTime time.Time) {
-	value := CachedTagSetList{
-		TagSets: tagsets,
-		Expiry:  startTime.Add(c.timeToLive),
+// fetchAndUpdateCachedTagSet updates the in-memory cache (asusming the update)
```
cuzelac: fix parens

metric_metadata/cached/cached.go
```diff
@@ -111,70 +158,121 @@ func (c *CachedMetricMetadataAPI) CheckHealthy() error {
...
+	defer item.Unlock()
 
-	c.addBackgroundGetAllTagsRequest(metricKey)
+	// Otherwise, we could be stale
+	if item.Stale.Before(c.clock.Now()) {
+		// Enqueue a background request
+		c.addBackgroundGetAllTagsRequest(item, metricKey)
+	}
+
+	defer context.Profiler.Record("CachedMetricMetadataAPI_Hit")()
```
cuzelac: move this record up to before `defer item.Unlock()`

cuzelac:
LGTM

PR #275, opened 2016-04-08
Return an empty channel on a nil Timeout

drcapulet:
Causing a nil pointer dereference error with an empty `TimeoutOwner`.

R: @Nathan-Fenner

### DIFF
--- a/tasks/canceller.go
+++ b/tasks/canceller.go
@@ -52,6 +52,10 @@ type Timeout struct {
 
 // Done returns a receive-only channel, which will close when the timeout ends.
 func (t *Timeout) Done() <-chan struct{} {
+	if t == nil {
+		return nil
+	}
+
 	return t.done
 }
 


Nathan-Fenner:
Receiving on a nil channel should block forever, per [Golang spec](https://golang.org/ref/spec#Receive_operator) so this change shouldn't be needed:

> For an operand ch of channel type, the value of the receive operation <-ch is the value received from the channel ch. The channel direction must permit receive operations, and the type of the receive operation is the element type of the channel. The expression blocks until a value is available. **Receiving from a nil channel blocks forever.** A receive operation on a closed channel can always proceed immediately, yielding the element type's zero value after any previously sent values have been received.

drcapulet:
So the issue is that `t *Timeout` itself is `nil`, so `t.done` tries to access a value on a nil pointer, causing the `nil pointer dereference error`.

I could easily return a nil channel instead of the `make` I'm currently using?

Nathan-Fenner:
Oh, right, that makes sense. I misread the condition.

I think it would be better to return `nil` here.

drcapulet:
Updated.

PR #276, opened 2016-04-18
use MakeFunction in place of explicit MetricFunction everywhere possible

Nathan-Fenner:
To try to make it easier to define `MetricFunction`s without having to write repetitive evaluations and casts everywhere, a new function `function.MakeFunction` (possible rename to just `Make`) has been added.

It asks for an `interface{}` which is a function that acts like a metric function. It can take any of:

- `string`
- `float64`
- `time.Duration`
- `api.SeriesList`
- `function.Value`
- `function.Expression`

In addition, it can have optional arguments by requesting pointers to any of these:

- `*string`
- `*float`
- `*time.Duration`
- `*api.SeriesList`
- `*function.Value`
- `*function.Expression`

Non-optional formal arguments cannot follow optional ones. (So, for example, `(float, *string, float)` would be illegal).

For all of these, the parameters will be evaluated before they are passed to the function except for `function.Expression` and `*function.Expression`, where it is up to the given function to evaluate them.

All evaluations are done **in parallel automatically**, so the function itself does not have to do any extra work here.

The argument function can also ask for an `function.EvaluationContext` to get the current context (or an `api.Timerange` to get just the current timerange).

If it wants to allow `group by` or `collapse by` clauses, it can ask for a `function.Groups` object. 

The argument function can return either `(function.Value, error)` or just a `function.Value` (which means it never errors). Either of these can be replaced by any subtype implementing these interfaces (e.g. `api.SeriesList` or `SyntaxError` respectively).

If an input function violates any of these constraints, `MakeFunction` panics.

TODO: test optional arguments, cleaning up reflection in `MakeFunction`.

TODO: variadic functions for `MakeFunction` (for variadic functions in MQE)

TODO: float, string, duration returns?

# Example

Consider the following toy metric function:

## Old Way

```
// adder takes one or two numbers.
// when given only one parameter, the second defaults to "1.0".
function.MetricFunction {
    Name: "adder",
    MinArguments: 1,
    MaxArguments: 2,
    Compute: func(context function.EvaluationContext, arguments []function.Expression, groups function.Groups) (function.Value, error) {
         left, err := function.EvaluateToDuration(arguments[0], context)
         if err != nil {
             return nil, err
         }
         right := float64(1.0)
         if len(arguments) == 2 {
             right, err = function.EvaluateToDuration(arguments[1], context)
         }
         return expression.Scalar(left + right), nil
    },
}
```

## New Way

```
// adder takes one or two numbers.
// when given only one parameter, the second defaults to "1.0".
function.MakeFunction(
    "adder",
    func(left float64, optionalRight *float64) float64 {
         right := float64(1.0)
         if optionalRight != nil {
             right = *optionalRight
         }
         return left + right
    },
)
```
in addition, the arguments will be evaluated in parallel here (there are library functions to evaluate them in parallel using the old way, but it will add several more lines for error-checking).

### DIFF
--- a/function/make.go
+++ b/function/make.go
@@ -209,9 +209,16 @@ func MakeFunction(name string, function interface{}) MetricFunction {
 			if len(output) == 2 && output[1].Interface() != nil {
 				return nil, output[1].Interface().(error)
 			}
-			if funcType.Out(0) == timeseriesType {
+			switch funcType.Out(0) {
+			case stringType:
+				return StringValue(output[0].Interface().(string)), nil
+			case scalarType:
+				return ScalarValue(output[0].Interface().(float64)), nil
+			case durationType:
+				return DurationValue{"", output[0].Interface().(time.Duration)}, nil
+			case timeseriesType:
 				return SeriesListValue(output[0].Interface().(api.SeriesList)), nil
-			} else {
+			default:
 				return output[0].Interface().(Value), nil
 			}
 		},
--- a/function/make.go
+++ b/function/make.go
@@ -17,6 +17,7 @@ package function
 import (
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/square/metrics/api"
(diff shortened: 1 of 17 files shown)


function/registry/registry.go
```diff
@@ -145,139 +142,63 @@ func MustRegister(fun function.MetricFunction) {
 
 // Constructor Functions
 
-// NewFilterCount creates a new instance of a count filtering function.
+// NewFilter creates a new instance of a filtering function.
 func NewFilterCount(name string, summary func([]float64) float64, ascending bool) function.MetricFunction {
```
drcapulet: Need to update either the method name of comment here

function/make.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	requiredArgumentCount := 0
+	optionalArgumentCount := 0
+	allowsGroupBy := false
+	for i := 0; i < funcType.NumIn(); i++ {
+		argType := funcType.In(i)
+		switch argType {
+		case contextType, timerangeType:
+		// Asks for part of context.
```
drcapulet: nit: indent

function/registry/registry.go
```diff
@@ -145,139 +142,63 @@ func MustRegister(fun function.MetricFunction) {
...
+				if *optionalDuration < 0 {
+					return api.SeriesList{}, fmt.Errorf("expected a positive duration but got %+v", *optionalDuration)
 				}
+				duration = *optionalDuration
 			}
-
-			if recentDuration < 0 {
-				return nil, fmt.Errorf("expected positive recent duration but got %+v", recentDuration)
+			if duration < 0 {
+				return api.SeriesList{}, fmt.Errorf("expected positive recent duration but got %+v", duration)
```
drcapulet: Probably don't need to check this twice? Can probably make the block above

```
if optionalDuration != nil {
    duration = *optionalDuration
}
```

query/command/command.go
```diff
@@ -273,9 +273,10 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (CommandResult, erro
 	case result := <-results:
 		lists := make([]api.SeriesList, len(result))
 		for i := range result {
-			lists[i], err = result[i].ToSeriesList(evaluationContext.Timerange, cmd.Expressions[i].QueryString())
-			if err != nil {
-				return CommandResult{}, err
+			var convErr *function.ConversionFailure
```
drcapulet: Why do you need to define this here?

Nathan-Fenner: You can't use `:=` when one of the things at the left is an index (like `lists[i]`).

drcapulet: Ah gotcha

function/registry/registry.go
Nathan-Fenner: fixed

function/make.go
Nathan-Fenner: fixed

function/registry/registry.go
Nathan-Fenner: fixed

drcapulet:
LGTM

PR #278, opened 2016-05-03
Move out `AddMetric(s)` methods from `MetricAPI`

Nathan-Fenner:
The majority of the code doesn't need to be able to `AddMetric(s)` to the metric API, since normal queries aren't allowed to modify the index (they can only read it).

"UI" has been renamed to "web" for clarity.

"Query" has been replaced with "console" for clarity, and no longer uses `liner`. This removes this dependency from MQE.


main/console/console.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+
+	"github.com/square/metrics/log"
+	"github.com/square/metrics/metric_metadata"
+	"github.com/square/metrics/metric_metadata/cached"
+	"github.com/square/metrics/metric_metadata/cassandra"
+	"github.com/square/metrics/query/command"
+	"github.com/square/metrics/query/parser"
+	"github.com/square/metrics/timeseries"
+
```
drcapulet: nit: these import lists should go together

main/console/console.go
```diff
@@ -0,0 +1,134 @@
...
+	go func() {
+		for range sigs {
+			pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
+		}
+	}()
+
+	config := struct {
+		ConversionRulesPath string           `yaml:"conversion_rules_path"`
+		Cassandra           cassandra.Config `yaml:"cassandra"`
+		Blueflood           blueflood.Config `yaml"blueflood"`
```
drcapulet: missing `:` after `yaml`

main/console/console.go
```diff
@@ -0,0 +1,134 @@
...
+
+	common.LoadConfigs(&config)
+
+	metadataAPI, err := cassandra.NewMetricMetadataAPI(config.Cassandra)
+	if err != nil {
+		common.ExitWithErrorMessage("Error loading Cassandra API: %s", err.Error())
+		return
+	}
+
+	// TODO: should Blueflood be responsible for this?
```
drcapulet: Responsible for loading the rules? It's a hard line to tow because ideally `blueflood.NewBlueflood` takes a config object that already have everything initialized.

Nathan-Fenner: Yeah, it's a weird case because some of Blueflood's config fields (tenant, host, etc.) are filled from YAML, but others have to be filled in afterward (namely the graphite converter).

The current situation is probably fine. I'll remove this comment.

main/console/console.go
```diff
@@ -0,0 +1,134 @@
...
+
+	config.Blueflood.GraphiteMetricConverter = &util.RuleBasedGraphiteConverter{Ruleset: ruleset}
+
+	blueflood := blueflood.NewBlueflood(config.Blueflood)
+
+	optimizedMetadataAPI := cached.NewMetricMetadataAPI(metadataAPI, cached.Config{
+		TimeToLive:   time.Minute * 5, // Cache items invalidated after 5 minutes.
+		RequestLimit: 500,
+	})
+	for i := 0; i < 10; i++ {
```
drcapulet: Do we really need the caching stuff in the bg for the console?

Nathan-Fenner: Probably not, that's a good point. I'll remove it (I just copied it over from the web main).

main/web/web.go
```diff
@@ -27,18 +26,19 @@ import (
 	"github.com/square/metrics/log"
 	"github.com/square/metrics/metric_metadata"
 	"github.com/square/metrics/metric_metadata/cached"
+	"github.com/square/metrics/metric_metadata/cassandra"
 	"github.com/square/metrics/query/command"
 	"github.com/square/metrics/timeseries"
+	"github.com/square/metrics/web"
 
 	"github.com/square/metrics/function/registry"
```
drcapulet: nit: these imports should go with the ones above

timeseries/blueflood/blueflood.go
```diff
@@ -66,6 +54,16 @@ type Config struct {
 	MaxSimultaneousRequests int `yaml:"simultaneous_requests"`
 }
 
+type BluefloodParallelRequest struct {
+	limit   int
+	tickets chan struct{}
+}
+
+type httpClient interface {
```
drcapulet: 😢 Really wish Go would wrap the stdlib in interfaces to allow for stuff like this instead of having to arbitrarily define types.

web/server.go
```diff
@@ -35,10 +36,14 @@ func NewMux(config Config, context command.ExecutionContext, hook Hook) *http.Se
 	httpMux.Handle("/token", tokenHandler{
 		context: context,
 	})
-	if config.JSONIngestion {
-		httpMux.Handle("/ingest", ingestHandler{
-			metricMetadataAPI: context.MetricMetadataAPI,
-		})
+	if updateAPI, ok := context.MetricMetadataAPI.(metadata.MetricUpdateAPI); ok && config.HTTPIngestion {
```
drcapulet: Maybe we should log in the case where `config.HTTPIngestion` is true but the `MetricMetadataAPI` isn't a `MetricUpdateAPI`?

web/server.go
```diff
@@ -35,10 +36,14 @@ func NewMux(config Config, context command.ExecutionContext, hook Hook) *http.Se
...
+		// If the API implements MetricUpdateAPI, the server also exposes an endpoint to ingest metrics.
+		// TODO: make it possible to disable this endpoint.
```
drcapulet: This TODO doesn't seem applicable anymore?

main/web/web.go
Nathan-Fenner: fixed

web/server.go
Nathan-Fenner: It might even be okay to go `Fatal`.

Right now one issue is that it's incompatible with the Caching API, but I can fix that with some interfaces.

drcapulet:
LGTM

Nathan-Fenner:
Just made some changes to the Cached API that are worth looking at before final merge.

metric_metadata/cached/cached.go
```diff
@@ -25,10 +25,21 @@ import (
 	"github.com/square/metrics/util"
 )
 
+// CachedAPI is a MetadataAPI that also supports background cache updates.
```
drcapulet: You'll want to update the comment here

metric_metadata/cached/cached.go
```diff
@@ -25,10 +25,21 @@ import (
...
+	metadata.MetricAPI
+	// GetBackgroundAction returns a function to be called to execute a background cache update.
+	GetBackgroundAction() func(metadata.Context) error
+	// CurrentLiveRequests returns the number of requests currently in the queue
+	CurrentLiveRequests() int
+	// MaximumLiveRequests returns the maximum number of requests that can be in the queue
+	MaximumLiveRequests() int
+}
+
 // CachedMetricMetadataAPI caches some of the metadata associated with the API to reduce latency.
```
drcapulet: Same here

drcapulet:
Few comments, otherwise LGTM

PR #281, opened 2016-05-18
Remove timerange from series list

Nathan-Fenner:
There is almost no benefit to storing the timerange of a series list in the list itself, since it will always be recorded in the associated `EvaluationContext` anyway.

Removing the `Timerange` field from `SeriesList` makes it easier to handle `SeriesList`s.

Only a few places were actually seriously effects:

* `ApplyTransformation` now has to ask for the resolution instead of finding it itself. This required only minor changes.
* `select_command_test.go` used a backend which needs `SeriesList`s to have timeranges in order to accurately query them. It's been replaced with `ComboAPI` (which has been modified slightly to support this).
* the JSON output is identical, but `QuerySeriesList` needed to be modified to report the timerange too.

@drcapulet

### DIFF
--- a/function/transform/transformation_test.go
+++ b/function/transform/transformation_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math"
 	"testing"
+	"time"
 
 	"github.com/square/metrics/api"
 	"github.com/square/metrics/function"
@@ -96,11 +97,10 @@ func TestTransformTimeseries(t *testing.T) {
 			}
 			ctx := function.EvaluationContext{}
 			seriesList := api.SeriesList{
-				Series:    []api.Timeseries{series},
-				Timerange: timerange,
+				Series: []api.Timeseries{series},
 			}
 
-			a, err := ApplyTransform(ctx, seriesList, transform.fun, params)
+			a, err := ApplyTransform(ctx, seriesList, transform.fun, params, timerange.Resolution())
 			result := a.Series[0]
 			if err != nil {
 				t.Error(err)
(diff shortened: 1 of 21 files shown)


function/transform/transformation_test.go
```diff
@@ -401,20 +375,15 @@ func TestApplyBound(t *testing.T) {
 		}
 	}
 	ctx := function.EvaluationContext{}
-	if _, err = ApplyTransform(ctx, list, Bound, []function.Value{function.ScalarValue(18), function.ScalarValue(17)}); err == nil {
+	if _, err := ApplyTransform(ctx, list, Bound, []function.Value{function.ScalarValue(18), function.ScalarValue(17)}, 30*time.Second); err == nil {
 		t.Fatalf("Expected error on invalid bounds")
 	}
-	if _, err = ApplyTransform(ctx, list, Bound, []function.Value{function.ScalarValue(-17), function.ScalarValue(-18)}); err == nil {
+	if _, err := ApplyTransform(ctx, list, Bound, []function.Value{function.ScalarValue(-17), function.ScalarValue(-18)}, 30*time.Second); err == nil {
```
drcapulet: not sure you need the `:=` here?

Nathan-Fenner: It wouldn't compile if the `err` was already in scope. Originally it was defined when creating the timerange for the test (which was removed).

drcapulet: Interesting, I must not have noticed it scoped it inside the if.

drcapulet:
LGTM 👍

PR #282, opened 2016-05-25
Fix Yaml Tag for Blueflood Config

Nathan-Fenner:
The default behavior for YAML decoding is to use the lowercase field name, so this doesn't actually cause any faulty behavior, but it should be fixed.

@drcapulet


drcapulet:
LGTM

PR #283, opened 2016-05-25
Reduce Usage of `api.ParseTagSet`

Nathan-Fenner:
Tag sets are unserialized in a lot of cases where they could just be literals.

@drcapulet


drcapulet:
LGTM

PR #284, opened 2016-05-25
Scalar Sets for Summary Functions

Nathan-Fenner:
Let's you do summarize timeseries into tagged scalars:

```
select
cpu | summarize.mean
from -1hr to now
```

or,

```
select
cpu | aggregate.sum(group by app) | summarize.mean
from -1hr to now
```

In addition, they play nice with joins, so you can write (for example):

```
select
cpu / (cpu | aggregate.sum(group by app) | summarize.current)
from -1hr to now
```

The UI currently cannot display the results of a scalar or tagged-scalar result.

@drcapulet

### DIFF
--- a/demo/example-config.yaml
+++ b/demo/example-config.yaml
@@ -14,7 +14,7 @@ blueflood:
 
 cassandra:
   hosts:
-    - localhost:9042                            # the IP addresses/hostnames for the Cassandra nodes (9160 is the default port, and can be omitted)
+    - localhost:9042                            # the IP addresses/hostnames for the Cassandra nodes
   keyspace: metrics_indexer                     # the keyspace for MQE indexing
 
 web:
--- a/demo/example-config.yaml
+++ b/demo/example-config.yaml
@@ -14,7 +14,7 @@ blueflood:
 
 cassandra:
   hosts:
-    - localhost:9160                            # the IP addresses/hostnames for the Cassandra nodes (9160 is the default port, and can be omitted)
+    - localhost:9042                            # the IP addresses/hostnames for the Cassandra nodes (9160 is the default port, and can be omitted)
   keyspace: metrics_indexer                     # the keyspace for MQE indexing
 
 web:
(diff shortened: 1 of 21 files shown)


function/scalarset.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/square/metrics/api"
+)
+
+type TaggedScalar struct {
+	TagSet api.TagSet
+	Value  float64
+}
+
+// TaggedScalar marshals NaN or infinity to null.
```
drcapulet: Comment here should be of the format `// MarshalJSON ...`

drcapulet:
Can you add tests for `summarize.last_not_nan` and `summarize.current` please? Otherwise LGTM.

Any specific reason you're removing the code for returning raw values?

Nathan-Fenner:
Will add tests.

I don't remember exactly why I decided to remove raw values in this PR. I think it was because I had to read through the custom JSON serializer for `Timeseries`. As far as I understand, `Raw` was only really ever used for debugging, and was off in all of our configurations. Not supporting it will make things a lot simpler.

I need to spin up a local instance of this to make sure the UI and JSON works correctly with scalars and timeseries.

Nathan-Fenner:
I found and fixed some issues with the UI/demo. I updated config files and the wiki to reflect some name changes, and fixed a pretty big bug in the startup functions.

For the moment, I'm not going to set up any UI to display scalar sets- there's just a message indicating that they're there and unviewable.

demo/example-config.yaml
```diff
@@ -14,7 +14,7 @@ blueflood:
 
 cassandra:
   hosts:
-    - localhost:9160                            # the IP addresses/hostnames for the Cassandra nodes (9160 is the default port, and can be omitted)
+    - localhost:9042                            # the IP addresses/hostnames for the Cassandra nodes (9160 is the default port, and can be omitted)
```
drcapulet: Mind fixing the comment here?

drcapulet:
Also need to remove the `emit` and `ingest` binaries.

PR #290, opened 2016-06-03
Change blueflood planning error condition

Nathan-Fenner:
This changes how Blueflood plans fetching. It has the following impacts:

* requesting data into the future is never a problem (Blueflood won't try to actually fetch the data, though)
* if the specified resolutions don't have monotonic (increasing) TTLs, then you'll unexpected get errors with queries. Skip resolutions with non-monotonic TTLs.

@drcapulet


drcapulet:
LGTM

PR #291, opened 2016-06-04
Profile individual requests

Nathan-Fenner:
Fixes a bug in profiling Blueflood (so now there will be per-fetch metrics).

@drcapulet


drcapulet:
LGTM

PR #294, opened 2016-06-09
Optimize Blueflood resolution selection loop into division

Nathan-Fenner:
There are two optimizations here:

* A for-loop has been replaced with a division in the interval-planning stage
* Interval-planning is performed only once total when fetching multiple series, and not once per timeseries

@drcapulet

### DIFF
--- a/timeseries/blueflood/blueflood.go
+++ b/timeseries/blueflood/blueflood.go
@@ -177,34 +177,20 @@ func planFetchIntervalsWithOnlyFiner(resolutions []Resolution, now time.Time, re
 	return planFetchIntervals(resolutions, now, requestRange.Interval())
 }
 
-// FetchSingleTimeseries fetches a timeseries with the given tagged metric.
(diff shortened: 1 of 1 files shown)


timeseries/blueflood/blueflood.go
```diff
@@ -136,21 +136,28 @@ func planFetchIntervals(resolutions []Resolution, now time.Time, requestInterval
...
+
+		// advance that number of intervals
+		newHere := here.Add(count * resolution.Resolution)
+
+		if newHere != here {
 			// At least one point is included, so:
-			answer[resolution] = api.Interval{Start: originalHere, End: here}
+			answer[resolution] = api.Interval{Start: here, End: newHere}
+			here = newHere
 		}
```
Nathan-Fenner: This is where the loop is optimized. The resulting code behaves _identically_.

drcapulet:
🚢

PR #296, opened 2016-06-14
Add error details to Blueflood

Nathan-Fenner:
When these errors happen, it's impossible to tell why. Now, details will be reported *and* MQE will respond with 500 instead of 400.

Giving up the structured errors isn't really a loss, since they're not being checked anywhere anyway.

@drcapulet

### DIFF
--- a/web/query.go
+++ b/web/query.go
@@ -220,6 +220,12 @@ func (q queryHandler) process(profiler *inspect.Profiler, parsedForm QueryForm) 
 	}, nil
 }
 
+// ErrorHTTP indicates that an error should override the return code.
+type ErrorHTTP interface {
+	error
+	ErrorCode() int
+}
+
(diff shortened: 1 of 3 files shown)


web/query.go
```diff
@@ -234,7 +240,11 @@ func (q queryHandler) ServeHTTP(writer http.ResponseWriter, request *http.Reques
 		}
 	default: // use the form parameters
 		if err := request.ParseForm(); err != nil {
-			writer.WriteHeader(http.StatusBadRequest)
+			if errHTTP, ok := err.(ErrorHTTP); ok {
```
drcapulet: I think you actually wanted the error around like 279.

Nathan-Fenner: Fixed; it now does this only for the error coming from `q.process`, which does the real work.

drcapulet:
🚢 on green!

PR #298, opened 2016-06-15
Blueflood HTTP Timeouts

Nathan-Fenner:
This PR changes the way that Blueflood handles timeouts. In particular:

* Blueflood's `Config` no longer has a `Timeout` field. Instead, you should set the timeout for a *query* (not just fetches to Blueflood) through the `Timeout` field in the `ExecutionContext`.
* Blueflood now respects the `Timeout` object passed by to it through `EvaluationContext` and cancels requests when it ends.
* Blueflood was organized slightly (mostly reordering functions).
* Testing support's `HTTPClient` now supports a `Do` function that just calls `Get` for mocking purposes.
* The `ParallelQueue` in `tasks` has been updated to use a passed-in `*tasks.Timeout` instead of creating its own from a `time.Duration`.

@drcapulet


timeseries/blueflood/blueflood.go
```diff
@@ -279,188 +267,48 @@ func (b *Blueflood) constructURL(metric api.TaggedMetric, interval api.Interval,
 	return result, nil
 }
 
-// performFetch is a synchronous method that fetches the given URL.
-func (b *Blueflood) performFetch(queryURL *url.URL) (queryResponse, error) {
-	resp, err := b.config.HTTPClient.Get(queryURL.String())
+type httpClient interface {
+	// our own client to mock out the standard golang HTTP Client.
+	Get(string) (resp *http.Response, err error)
```
drcapulet: No need to name the return parameters here.

drcapulet:
Overall LGTM with one comment, any thoughts on using the `golang.org/x/net/context` instead of our timeout solution?

Nathan-Fenner:
I think using the `net/context` instead of `Timeout` would be a good idea.

drcapulet:
LGTM, let's switch to `net/context` in a follow up PR.

PR #300, opened 2016-06-20
define tag.copy

Nathan-Fenner:
Defines the `tag.copy` function that behaves thusly:

```
tag.copy(series, target, source)
```
will return a clone of `series` where the tag `target` will have the value of `source` for each series. For example,

Input:
* Series
  * [1,2,3,4] `app:mqe` `host:server1`
  * [5,6,7,8] `app:mqe` `host:server2`
  * [1,2,4,8] `app:indexer` `host:server3`
  * [2,3,5,7] `app:indexer` `host:server4`
* Target: `service`
* Source: `app`

Output:
* [1,2,3,4] `app:mqe` `host:server1` `service:mqe`
* [5,6,7,8] `app:mqe` `host:server2` `service:mqe`
* [1,2,4,8] `app:indexer` `host:server3` `service:indexer`
* [2,3,5,7] `app:indexer` `host:server4` `service:indexer`

It can be used to improve agreement for joins like `-` or `/`, as well as to improve display (by renaming a tag, via `... | tag.copy(new, old) | tag.drop(old)`.

If the `source` tag is absent in the original, `target` will be deleted (not just set to `""`) (even if present). Is this the best behavior? Other possible behaviors would be to set it to `""`, to leave it alone, or to throw an error.

@drcapulet

### DIFF
--- a/api/tagset.go
+++ b/api/tagset.go
@@ -21,11 +21,23 @@ import (
 )
 
 // TagSet is the set of key-value pairs associated with a given metric.
+// Instances of tag set should generally be treated immutably to avoid
+// accidentally modifying instances belonging to other metrics.
 type TagSet map[string]string
 
 // NewTagSet creates a new instance of TagSet.
 func NewTagSet() TagSet {
 	return make(map[string]string)
+}
+
+// Clone returns a new, identical tag set to the method receiver. It's safe to
+// mutate until you give it away.
+func (tagSet TagSet) Clone() TagSet {
+	result := NewTagSet()
+	for key, value := range tagSet {
+		result[key] = value
+	}
+	return result
 }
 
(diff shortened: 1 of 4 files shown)


drcapulet:
I think it should probably throw an error if the the tag you asked to copy doesn't exists & if the destination is empty, WDYT?

Nathan-Fenner:
Right now, MQE doesn't really enforced any kind of "tag schema" on metrics, so there might be one line with a tag, and another without that same tag. So I think throwing an error might be too extreme, just it might only affect a small number of metrics.

drcapulet:
Fair enough, though I think erroring with an empty destination is better - we've got `tag.delete` if you want to explicitly delete a tag.

drcapulet:
LGTM on green.

PR #302, opened 2016-06-23
Memoize Function Evaluations

Nathan-Fenner:
`function.EvaluationContext` is made immutable (by virtue of keeping its fields private) so that it can be reliably left untouched by metric functions. This introduces some boilerplate, but it's mostly just getter-methods for the `EvaluationContext` type.

`QueryString() string` and `Name() string` for the `Expression` interface have been replaced by `ExpressionString(DescriptionMode) string` that handles both of these *and* memoization.

There are now two kinds of expressions: `Expression` and `ActualExpression`.

The first is the kind that's used in evaluation (for example, functions should expect their arguments to be `Expression`):
```
type Expression interface {
    Evaluate(EvaluationContext) (Value, error)
    ExpressionString(DescriptionMode) string
}
```

The second is what expressions should (generally) be implemented as:
```
type ActualExpression interface {
     ActualEvaluate(EvaluationContext) (Value, error)
     ExpressionString(DescriptionMode) string
}
```

To convert an `ActualExpression` to an `Expression`, use `function.Memoize(ActualExpression) Expression`.

Memoization **relies on `StringExpression` being implemented correctly.** Distinct expressions **must** have distinct names, or they will be erroneously conflated during evaluation.

## `transform.alias` is now fully deprecated

Renaming functions by introspecting their arguments is not pleasant; this functionality has been removed instead of making things complicated.

The function remains defined and is a no-operations (other than adding a note indicating that it's deprecated). It no longer has any affect on metric names.

## Limitations

The main limitation of this memoization is that it doesn't apply to different subcontexts that are actually identical. The most obvious issue is in an expression like

```
select
transform.timeshift(cpu, -1w) + transform.timeshift(cpu, -1w)
from -30m to now
```

where the two fetches happen in distinct subcontexts and hence cannot be shared (even though they happen to be the same).

### DIFF
--- a/function/builtin/transform/special.go
+++ b/function/builtin/transform/special.go
@@ -26,7 +26,7 @@ import (
 var Timeshift = function.MakeFunction(
 	"transform.timeshift",
 	func(expression function.Expression, duration time.Duration, context function.EvaluationContext) (function.Value, error) {
-		newContext := context.WithTimerange(context.Timerange.Shift(duration))
+		newContext := context.WithTimerange(context.Timerange().Shift(duration))
 		return expression.Evaluate(newContext)
 	},
 )
@@ -35,15 +35,14 @@ var MovingAverage = function.MakeFunction(
 	"transform.moving_average",
 	func(context function.EvaluationContext, listExpression function.Expression, size time.Duration) (api.SeriesList, error) {
 		// Applying a similar trick as did TimeshiftFunction. It fetches data prior to the start of the timerange.
-		limit := int(float64(size)/float64(context.Timerange.Resolution()) + 0.5) // Limit is the number of items to include in the average
+		limit := int(float64(size)/float64(context.Timerange().Resolution()) + 0.5) // Limit is the number of items to include in the average
 		if limit < 1 {
 			// At least one value must be included at all times
 			limit = 1
 		}
 
-		timerange := context.Timerange
+		timerange := context.Timerange()
 		newTimerange := timerange.ExtendBefore(time.Duration(limit-1) * timerange.Resolution())
-
 		newContext := context.WithTimerange(newTimerange)
 		// The new context has a timerange which is extended beyond the query's.
 		list, err := function.EvaluateToSeriesList(listExpression, newContext)
@@ -54,7 +53,7 @@ var MovingAverage = function.MakeFunction(
 		// Update each series in the list.
 		for index, series := range list.Series {
 			// The series will be given a (shorter) replaced list of values.
-			results := make([]float64, context.Timerange.Slots())
+			results := make([]float64, context.Timerange().Slots())
 			count := 0
 			sum := 0.0
 			for i := range series.Values {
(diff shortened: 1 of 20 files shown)


function/expression.go
```diff
@@ -35,7 +35,7 @@ import (
 // from the backend systems.
 // * Contains current timerange being queried for - this can be
 // changed by say, application of time shift function.
-type EvaluationContext struct {
+type EvaluationContextBuilder struct {
```
drcapulet: Need to update the comment here.

function/expression.go
```diff
@@ -123,13 +177,25 @@ func (c FetchCounter) Consume(n int) error {
 // timeseries according to the resolution specified in its EvaluationContext's
 // Timerange. Internal nodes may assume that results from evaluating child
 // Expressions correspond to the timerange in the current EvaluationContext.
-type Expression interface {
+type ActualExpression interface {
```
drcapulet: Need to update comments here.

function/make.go
```diff
@@ -227,5 +229,6 @@ var expressionType = reflect.TypeOf((*Expression)(nil)).Elem()
 var groupsType = reflect.TypeOf(Groups{})
 var contextType = reflect.TypeOf(EvaluationContext{})
 var timerangeType = reflect.TypeOf(api.Timerange{})
+var evaluationNotesType = reflect.TypeOf(EvaluationContext{})
```
drcapulet: Seems weird that they ask for notes and get an `EvaluationContext `, why not `EvaluationNotes`?

function/builtin/transform/special.go
```diff
@@ -140,8 +140,8 @@ var ExponentialMovingAverage = function.MakeFunction(
 
 // Alias is deprecated.
```
drcapulet: Let's just remove this completely?

function/expression.go
Nathan-Fenner: Fixed; also added some more comments and replaced `NewEvaluationContext(builder)` with `builder.Build()` for ergonomics.

function/expression.go
Nathan-Fenner: Fixed

function/make.go
Nathan-Fenner: Whoops- this was used as an untested convenience method. I just removed it because I couldn't see enough use for it, and it poses a little bit of a challenge since you need to request it as a pointer to be useful, which would probably confuse `MakeFunction`.

function/builtin/transform/special.go
Nathan-Fenner: Okay- any queries using it will have to be updated, because they'll start to error

function/builtin/transform/special.go
```diff
@@ -88,11 +87,13 @@ var ExponentialMovingAverage = function.MakeFunction(
 	"transform.exponential_moving_average",
 	func(context function.EvaluationContext, listExpression function.Expression, size time.Duration) (api.SeriesList, error) {
 		// Applying a similar trick as did TimeshiftFunction. It fetches data prior to the start of the timerange.
-		limit := int(float64(size)/float64(context.Timerange.Resolution()) + 0.5) // Limit is the number of (additional) items to include in the average
-
-		timerange := context.Timerange
+		limit := int(float64(size)/float64(context.Timerange().Resolution()) + 0.5) // Limit is the number of items to include in the average
+		if limit < 1 {
```
drcapulet: Is there a test to check this?

Nathan-Fenner: There wasn't: I just added some, and also made the moving average functions error if they're given a negative duration (0 is fine and leaves the input unchanged).

drcapulet:
LGTM

PR #303, opened 2016-06-24
don't omit series when empty

Nathan-Fenner:
This fixes a minor UI bug in MQE, and possibly broken interface when talking to other services over JSON.

@drcapulet


drcapulet:
LGTM

PR #308, opened 2016-06-29
Replace timeout with context

Nathan-Fenner:
Replaces `tasks.Timeout` with `context.Context` from `golang.org/x/net/context`.

@drcapulet


function/builtin/transform/transformation_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"math"
 	"testing"
 
+	"golang.org/x/net/context"
```
drcapulet: nit: lets sort imports as:

```
import (
  stdlib

  project code

  external code
)
```

query/command/command.go
```diff
@@ -272,9 +276,8 @@ func (cmd *SelectCommand) Execute(context ExecutionContext) (CommandResult, erro
 		results <- result
 	}()
 	select {
-	case <-timeout:
-		return CommandResult{}, function.NewLimitError("Timeout while executing the query.",
-			context.Timeout, context.Timeout)
+	case <-ctx.Done():
+		return CommandResult{}, function.NewLimitError("Timeout while executing the query.", context.Timeout, context.Timeout)
```
drcapulet: Any reason we pass `context.Timeout` twice here?

Nathan-Fenner: It's "actual" and "limit" which are the same here (since we don't go over by any meaningful amount, we don't record the actual time)

function/builtin/transform/transformation_test.go
Nathan-Fenner: Sorted now

drcapulet:
LGTM

PR #310, opened 2016-06-30
With context and better sharing for memoization

Nathan-Fenner:
Add a `WithAdditionalConstraint(predicate.Predicate)` method to `EvaluationContext` that plays nice with memoization.

Also, add a `memoizationMap` to the `EvaluationContext` which can be used to look up already-existing `memoization`s within the same query, which makes the following query much more efficient, by performing the fetch for `cpu` only once instead of twice:

```
transform.timeshift(cpu, -1w) + transform.timeshift(cpu, -1)
from -30m to now
```

### DIFF
--- a/function/memoize.go
+++ b/function/memoize.go
@@ -89,3 +89,24 @@ func (m memoizedExpression) Evaluate(context EvaluationContext) (Value, error) {
 func (m memoizedExpression) ExpressionString(mode DescriptionMode) string {
 	return m.Expression.ExpressionString(mode)
 }
+
+////////
+
+type memoizationMap struct {
+	sync.Mutex
+	Map map[contextIdentity]*memoization
+}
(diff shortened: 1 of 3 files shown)


function/memoize.go
```diff
@@ -89,3 +89,23 @@ func (m memoizedExpression) Evaluate(context EvaluationContext) (Value, error) {
...
+type memoizationMap struct {
+	sync.Mutex
+	Map map[contextIdentity]*memoization
+}
+
+func (m *memoizationMap) get(i contextIdentity) *memoization {
+	m.Lock()
+	defer m.Unlock()
+	_, ok := m.Map[i]
+	if !ok {
```
drcapulet: You can just do `if _, ok := m.Map[i]; !ok {`

PR #321, opened 2016-08-12
Inspect query to adjust timerange

Nathan-Fenner:
When doing a query like

```
select
transform.moving_average(foo, 5m)
from -1d to now
```

when the edge of availability of full resolution data is at `-1d`, since you'd fetch data beyond that point with the moving average, the query would suddenly error.

Now, by inspecting the query (using annotations on the registered functions) it's possible to automatically adjust the query's resolution, increased to (say) `5m`, so that no error occurs.

It's intelligent enough to recognize that

```
select
transform.moving_average(foo, 5m) | transform.moving_average(12m)
from -1d to now
```

extends by a total of 17m, and that

```
select
transform.moving_average(foo, 5m) | transform.timeshift(-7m)
from -1d to now
```

doesn't extend it at all.

@drcapulet

### DIFF
--- a/function/builtin/forecast/rolling_function.go
+++ b/function/builtin/forecast/rolling_function.go
@@ -63,6 +63,7 @@ var FunctionRollingMultiplicativeHoltWinters = function.MakeFunction(
 
 		return result, nil
 	},
+	function.Option{Name: "WidenBy", Value: function.Argument(5)},
 )
 
 // FunctionRollingSeasonal is a forecasting MetricFunction that performs the rolling seasonal estimation.
@@ -104,6 +105,7 @@ var FunctionRollingSeasonal = function.MakeFunction(
 
 		return result, nil
 	},
+	function.Option{Name: "WidenBy", Value: function.Argument(3)},
 )
 
 // FunctionLinear forecasts with a simple linear regression.
@@ -140,4 +142,5 @@ var FunctionLinear = function.MakeFunction(
 
 		return result, nil
 	},
+	function.Option{Name: "WidenBy", Value: function.Argument(1)},
 )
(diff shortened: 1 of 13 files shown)


function/builtin/forecast/rolling_function.go
```diff
@@ -63,6 +63,7 @@ var FunctionRollingMultiplicativeHoltWinters = function.MakeFunction(
 
 		return result, nil
 	},
+	function.Option{Name: "WidenBy", Value: function.Argument(5)},
```
drcapulet: Can we make the `Name` and enum?

function/make.go
```diff
@@ -215,6 +223,49 @@ func MakeFunction(name string, function interface{}) MetricFunction {
...
 	}
+
+	for _, option := range options {
+		// TODO: check there aren't multiple conflicting values
+		if option.Name == "WidenBy" || option.Name == "ShiftBy" {
+			sign := -1
+			if option.Name == "ShiftBy" {
+				sign = 1
+			}
+			switch value := option.Value.(type) {
```
drcapulet: Let's and an `else` clause here to fail if it doesn't know how to handle `option.Value.(type)`

function/builtin/forecast/rolling_function.go
Nathan-Fenner: Done.

function/make.go
Nathan-Fenner: It's in a switch; there's now a `default` that panics (`MakeFunction` panics whenever something goes wrong, because it's only intended to be called by the programmer as MQE is starting up).

Nathan-Fenner:
Won't work on `forecast.anomaly_*` functions with extra time.

PR #329, opened 2017-02-15
Autocomplete UI fixes

DanielNi:
- Dispatch a new "input" event so that autocomplete actually works
- Explicitly set height of the first tab so autocomplete options are no longer hidden

Still need to fix the broken build...

@syamp @drcapulet


syamp:
My suggestion it to vendor the parser library from the last known version that we use. Otherwise lgtm

DanielNi:
Rebased off master, which fixed the build.

PR #331, opened 2017-06-20
Update Go versions and vendor using Godep (and fix the build)

DanielNi:
- Vendor latest versions of dependencies using Godep.
- Update supported Go versions and move to the native "context" package
- Regenerates the language using the latest version of peg
- Don't check licenses in vendor/
- Workaround to fix a Cassandra startup bug in Travis CI

All of this to get the build to pass...


query/command/command.go
```diff
@@ -15,6 +15,7 @@
 package command
 
 import (
+	netcontext "context"
```
drcapulet: I think we could probably remove the `netcontext` here?

DanielNi:
@drcapulet I guess we could import the peg library somewhere (maybe in `parser.go`?), otherwise I don't think it'll work with Godep.

bwester:
Godep will only vendor imported packages, not commands. I wouldn't want to roll something on our own here just to get a stable copy of `peg`. Checking in generated source code will work well enough.

That Travis CI build issue you mentioned is from 2015. How has this project not run into this problem before?

DanielNi:
Looking at previous builds, it looks like Cassandra has always failed to start up, but the tests still passed somehow. Maybe something in the newer versions of Go causes the tests to fail.

